PR #52, opened 2015-01-06
NO_CONTRACTS documenting

kinnrot:
added notes about no contracts


egonSchiele:
Hi kinnrot,
The README is not the correct place for this. I have added it to the tutorial.

Issue #51, opened 2015-01-01
Make this gem development/test environments only


kinnrot:
No reason to let this code run in production, is this easy to make?

egonSchiele:
Check out the section on failure callbacks here:
http://egonschiele.github.io/contracts.ruby/

You can just disable the check for prod envs.

kinnrot:
It allows me to ignore a failure, but not disable the check, it would be
nice if I could use this in develop with zero performance impact in
production...

alex-fedorov:
@kinnrot This gem allows you to do a technique called pattern matching: define your method multiple times with different contracts and it will call the first one that matches. Which is a neat feature that Ruby itself lacks. If you disable this gem for production by default => that means that code using this feature will break in production.

alex-fedorov:
@kinnrot if you need a fast solution with zero impact in production, you can use this simple snippet:

```ruby
if FrameworkOfChoice.production_env?
  module Contracts
    def self.common(base)
      base.class_eval do
        def Contract(*args)
          # do nothing for production
        end
      end
    end
  end
end
```

kinnrot:
Thanks !!! Great answers

egonSchiele:
I've added support for a new environment flag in master with this commit. So if you run your code with the NO_CONTRACTS environment variable set, the contracts won't be checked at all.

alex-fedorov:
@egonSchiele I ended up implementing custom stuff for pattern matching using code like that

```ruby
Match Contract None => nil
def smthing
  nil
end

Match Contract Num => Num
def smthing(x)
  42
end
```

Which allows me to fix the problem with so-called "deep" pattern matching (whenever you have a contract violation somewhere deeper in your method calls, it will treat it the same and proceed to next pattern and succeed, which should not be like that (usually) ).

In case of having this explicit `Match` I have implemented in a way so it captures a next defined method (like `contracts.ruby`, ie this gem, does) and wraps it in a begin-rescue block which catches all inner `ContractError`-s and re-raises them with different exception type, which defeats the problem of "deep" pattern matching (which is still can be used if needed without `Match` invocation).

After that I was able to write a small module `MyNamespace::Contracts` which on inclusion will include ruby.contracts' `::Contracts` and use `super` to call original `Contract(*args)` method if forced. This "forced" can depend on multiple conditions, in my case it for example smth like that:

- if special environment variable is not set => use `super` always
- otherwise
[...]

PR #53, opened 2015-01-07
Bug: Eliminates problems with pattern matching with deep contract violations

alex-fedorov:
related to this comment https://github.com/egonSchiele/contracts.ruby/issues/51#issuecomment-68742587

@egonSchiele I have added a fix for deep contract violations when pattern matching. I was able to do that without changing public api (i.e. no additional `Match` keyword or anything like that).


egonSchiele:
Merged, thanks!

PR #57, opened 2015-01-09
Cover deep contract violation during pattern matching better

waterlink:
Explanation here: https://github.com/waterlink/contracts.ruby/commit/0f8287dc31af81ee067f573e5529d9b3ac0dedd6#commitcomment-9225203

### DIFF
--- a/spec/fixtures/fixtures.rb
+++ b/spec/fixtures/fixtures.rb
@@ -86,7 +86,7 @@ def person(data)
 end
 
 Contract Proc => Any
-def call(&blk)
+def do_call(&blk)
   blk.call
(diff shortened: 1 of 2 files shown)


spec/fixtures/fixtures.rb
```diff
@@ -86,7 +86,7 @@ def person(data)
 end
 
 Contract Proc => Any
-def call(&blk)
+def do_call(&blk)
```
waterlink: Not the best idea to define `call` on `Object`. It mess with `rspec-mocks`, which makes it run into stack level too deep.

PR #62, opened 2015-01-13
Bug/37 :: Allow usage of singleton classes

### LINKED ISSUE #37: Singleton class doesn' work?
Hi!

Don't work contracts.rb in like this  singleton class case. Please advice me.
<pre>
class ContractsTest
  class &lt;&lt; self
    Contract String => None
    def hoge(str)
      p str
    end
  end
end
</pre>

not check param type

<pre>
ContractsTest.hoge 10
</pre>

> output: 10

waterlink:
fixes #37 

/cc @egonSchiele

### DIFF
--- a/spec/fixtures/fixtures.rb
+++ b/spec/fixtures/fixtures.rb
@@ -276,6 +276,11 @@ class MyBirthday < Struct.new(:day, :month)
 end
 
 class SingletonClassExample
+  # This turned out to be required line here to make singleton classes
+  # work properly under all platforms. Not sure if it worth trying to
+  # do something with it.
+  include Contracts
+
   class << self
     Contract String => String
     def hoge(str)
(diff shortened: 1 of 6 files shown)


waterlink:
Phew, it was a tough one.

spec/fixtures/fixtures.rb
```diff
@@ -274,3 +274,17 @@ def next_year!
     self.day = 1
   end
 end
+
+class SingletonClassExample
+  # This turned out to be required line here to make singleton classes
+  # work properly under all platforms. Not sure if it worth trying to
+  # do something with it.
+  include Contracts
```
waterlink: I was forced to use an inclusion here regardless of `Contracts` being included already in `Object` (or somewhere in hierarchy) or not. Because using this is the only portable way to link eigenclass with its owner class.

I don't really know if it worth to try to get rid of it. I would rather fail with descriptive error when somebody try to use contracts inside of singleton class without making this inclusion first in original class.

waterlink: /cc @egonSchiele

egonSchiele:
Tough one for sure o.O

PR #67, opened 2015-01-17
[DONT MERGE, opinion required] Bug/66: Self in inherited singleton methods

### LINKED ISSUE #66: `self` in inherited singleton methods gives parent class
If I put a contract on a singleton method, when inherited self refers to it's parent class, rather then itself.
Here's a rough spec if it helps clarify what I mean:

``` ruby
class A
  include Contracts

  def self.my_normal_method
    self
  end

  Contract Any => Any
  def self.my_contracted_method
    self
  end
end

class B < A
end

describe B do
  it do
    expect(B.my_normal_method).to eq(B)
  end

  it do
    expect(B.my_contracted_method).to eq(B)
  end
end
```

The second test would fail.

waterlink:
fixes #66

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -267,6 +267,7 @@ class Contract < Contracts::Decorator
       end
     end
 
+    # NOTE: each respond_to? check takes some time, actually.. Why not precalculate type of @method?
     result = if @method.respond_to? :bind
       # instance method
       @method.bind(this).call(*args, &blk)
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -271,11 +271,12 @@ class Contract < Contracts::Decorator
     result = if @method.respond_to? :bind
       # instance method
       @method.bind(this).call(*args, &blk)
-    elsif @method.respond_to?(:unbind)
-      # class method
+    elsif @method.respond_to?(:unbind) && RUBY_VERSION.to_f > 1.8
+      # class method for ruby 1.9+
       @method.unbind.bind(this).call(*args, &blk)
     else
       # Proc, lambda, block
+      # or class method for ruby 1.8
       @method.call(*args, &blk)
     end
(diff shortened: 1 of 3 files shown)


waterlink:
Looks like this couldn't be done in ruby 1.8.. Looking if there is a way for workaround

waterlink:
So, I tried out all possible other ways, like doing the same but at the singleton class level, or using `to_proc`, but all of them fail in 1.8.7 with the same error.

So the error is:

```
TypeError:
       singleton method bound for a different object
```

I see only 3 solutions:
- accept this bug as wontfix
- drop support for 1.8
- guard this feature for 1.8, meaning that this bug will be fixed for 1.9+ only, but will be present on 1.8. And of course add this special case somewhere to documentation or FAQ, stating that strange behavior for 1.8.

@egonSchiele WDYT?

waterlink:
Other solution, is to change the way we apply decorator to class. Currently we are redefining the method, storing reference of original method. Instead of just storing reference of original method, we could for example use `alias_method` to change its name to something unique to `contracts.ruby`, i.e.:

```ruby
Contract ...
def my_method(...)
```

in the end becomes:

```ruby
def my_method(...)
  # original implementation untouched
end

alias_method :__contracts_ruby_original_my_method_0, :my_method

def my_method(...)
  # checks contract and calls :__contracts_ruby_original_my_method_0
end
```

If you notice `_0` at the end of method name, it is to support current implementation of pattern matching (or method overloading): for other patterns (overloads) it should be `_1`, `_2` and so on.

That way we wouldn't even run into this nasty bug with `self` being unexpected.

WDYT? @egonSchiele

waterlink:
Actually last solution could even speed up performance, but I'm not entirely sure.

waterlink:
Closing in favour of #70

PR #68, opened 2015-01-17
Bug/65: Args splat with block argument at the end

### LINKED ISSUE #65: *args, &block
Hi, great gem!

Unless I've missed something, it seems that for methods with both a splat and a block, the block gets put in with the splat for the contract checks. This is the only way I've found to get it too work:

``` ruby
Contract Args[Or[Symbol, Proc]] => Any
def method(*args, &b)
```

Which allows procs to be put in any where in the splat. Is there a better way of doing this? If so, can it be added as an example in the tutorial?

Thanks.

waterlink:
fixes #65

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -262,7 +262,7 @@ class Contract < Contracts::Decorator
       # contracts on methods
       @args_contracts.each_with_index do |contract, i|
         if contract.is_a? Contracts::Func
-        args[i] = Contract.new(@klass, args[i], *contract.contracts)
+          args[i] = Contract.new(@klass, args[i], *contract.contracts)
         end
       end
     end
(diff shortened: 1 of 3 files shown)


egonSchiele:
Merged, but the error message is a little inaccurate. Suppose I have this code:

```ruby
Contract Num, Proc => nil
def double x, &blk
  puts x * 2
  blk.call
  nil
end

double(4)
```

The error says:

```
    Expected: Proc,
    Actual: 4
    Value guarded in: Object::double
    With Contract: Num, Proc => NilClass
```

`Actual: 4` is wrong, should be `Actual: nil` or something.

egonSchiele:
Also, with this change, you can't specify a contract on the block. i.e. this doesn't work:

```ruby
Contract Num, Func[Num => Num] => nil
def double x, &blk
  blk.call(x)
  nil
end
```

waterlink:
Hmm, I can see a fix for `Func`. About inaccurate message will try to do something to work it around.

PR #70, opened 2015-01-19
Bug/66: fix self in inherited singleton methods by changing original method preserving strategy

### LINKED ISSUE #66: `self` in inherited singleton methods gives parent class

waterlink:
fixes #66 

As I thought, changing to aliasing instead of preserving method object, saves the day. Not only fixes the bug, but improves performance nicely:

master:

```
testing add                      0.460000   0.000000   0.460000 (  0.464251)
testing contracts add            4.380000   0.020000   4.400000 (  4.403810)
```

this branch:

```
testing add                      0.450000   0.010000   0.460000 (  0.448705)
testing contracts add            3.530000   0.010000   3.540000 (  3.542390)
```

@egonSchiele can you confirm performance improvement?

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -1,5 +1,6 @@
 require 'contracts/core_ext'
 require 'contracts/support'
+require 'contracts/method_reference'
 require 'contracts/errors'
 require 'contracts/eigenclass'
 require 'contracts/decorators'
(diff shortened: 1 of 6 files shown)


waterlink:
updated to be merge-able

egonSchiele:
Performance improvement confirmed. Old:

```
testing add                      0.360000   0.000000   0.360000 (  0.360998)
testing contracts add            4.810000   0.010000   4.820000 (  4.808664)
```

New:

```
testing add                      0.360000   0.000000   0.360000 (  0.369388)
testing contracts add            4.390000   0.010000   4.400000 (  4.387795)
```

PR #71, opened 2015-01-20
NO_CONTRACTS mode :: Allow pattern matching to work when contracts are disabled

waterlink:
Allows to:

- Use contracts everywhere for the sake of guarantees - run in test/dev/staging mode
- Benefit from zero-overhead `NO_CONTRACTS=yes` in production
- Benefit from pattern matching (aka method overloading) on all environments


egonSchiele:
Makes sense, thanks.

PR #73, opened 2015-01-21
fixes #72: a bit clever handling of splats and blocks combinations

### LINKED ISSUE #72: Should allow contracts on blocks
Issue for my comment in pull #68:

Currently, I can't add a contract on a block like this:

```ruby
Contract Num, Func[Num => Num] => nil
def double x, &blk
  blk.call(x)
  nil
end
```

waterlink:
fixes #72 
followup to #68 

Now it properly handles different combinations of presence of: splat in contract, proc/func/method in contract and block in arguments.

/cc @egonSchiele

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -239,33 +239,61 @@ class Contract < Contracts::Decorator
   end
 
   def call_with(this, *args, &blk)
-
     _args = blk ? args + [blk] : args
+
+    size = @args_contracts.size
+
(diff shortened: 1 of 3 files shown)


waterlink:
BTW, whenever I try to refactor a bit `call_with` method, it causes huge slowdown immediately :(

waterlink:
@egonSchiele any news on this one?

egonSchiele:
oops, thought I merged this already...

PR #78, opened 2015-02-04
Bug/77 :: fix constant scoping issues

### LINKED ISSUE #77: Constant scoping issues
This looks like a very promising lib, but I get `NameError`'s when a contract is violated.
I'm on ruby 2.1.2.

Using the version from rubygems.org (0.5), I get this error:

    contracts-0.5/lib/contracts.rb:134:in `failure_msg': uninitialized constant Contract::Support (NameError)

Using master (26de665a), I get this error:

    lib/contracts.rb:247:in `call_with': uninitialized constant Contract::Func (NameError)

It looks like the `Contract` class is trying to access constants inside `Contracts` (plural) without specifying the scope via `::`. Is there a stable version I can use?

Also, I think it would be helpful to have tags & branches for each release so it's easier to view changes between releases.

waterlink:
fixes #77 

- Removes global inclusion of `Contracts` in `Object` from specs, so that there is no nasty bugs.

### DIFF
--- a/spec/contracts_spec.rb
+++ b/spec/contracts_spec.rb
@@ -1,8 +1,6 @@
-include Contracts
-
 RSpec.describe "Contracts:" do
   before :all do
-    @o = Object.new
+    @o = GenericExample.new
   end
 
   describe "basic" do
--- a/spec/contracts_spec.rb
+++ b/spec/contracts_spec.rb
@@ -84,12 +84,13 @@ RSpec.describe "Contracts:" do
 
     context "when owner class does not include Contracts" do
       let(:error) {
-        # NOTE Unable to support this user-friendly error for 1.8.7
-        # it has much less support for singleton inheritance hierarchy
-        if RUBY_VERSION.to_f < 1.9
-          [NoMethodError, /undefined method `Contract'/]
-        else
+        # NOTE Unable to support this user-friendly error for ruby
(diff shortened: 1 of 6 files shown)


spec/contracts_spec.rb
```diff
@@ -85,9 +83,20 @@
     end
 
     context "when owner class does not include Contracts" do
+      let(:error) {
+        # NOTE Unable to support this user-friendly error for ruby
```
waterlink: So eigenclasses in jruby and ruby 1.8.7 are really retarded.. They don't have even hierarchy. So it is not possible to have `Contract` method without calling `include Contracts` on owner of eigenclass.

/cc @egonSchiele 

WDYT, is it okayish fix for spec failure or not.

Basically what changes between normal versions and these ones is that user will not get nice error when he forgot to include `Contracts` in owner class of eigenclass. He will just get `NoMethodError`.

waterlink:
@egonSchiele Can you look into this one?

spec/contracts_spec.rb
waterlink: I tried to approach the problem with `inherited` hook, but it failed, because if you have no method error in class definition body in 1.8.7, then class will not be defined at all, and hook will not fire.

egonSchiele:
Looks good. I'm ok with showing a NoMethodError message in ruby 1.8.7/jruby for now.

PR #84, opened 2015-02-05
fixes #79 :: mention "pattern matching"

### LINKED ISSUE #79: Tutorial: overloading == pattern matching
I don't know if you realize this, but your library implements [pattern matching](http://en.wikipedia.org/wiki/Pattern_matching). One thing "missing" is that you can't automatically pass the matched values as local variables to the method body (but I'm not even sure if this is essential to PM).

Example: I can specify arbitrary nesting

    Contract ({a: Num, b: [[String]]}) => String
    def add h
      h[:a].to_s + h[:b][0][0]
    end
    add(a: 2, b: [['4']])

The tutorial calls this "overloading", but I think it is also useful to use the name "pattern matching" so Googling for "pattern matching in ruby" would lead here :)

kliuless:


TUTORIAL.md
```diff
@@ -434,6 +436,20 @@ end
 
 For an argument, each function will be tried in order. The first function that doesn't raise a `ContractError` will be used. So in this case, if x == 1, the first function will be used. For all other values, the second function will be used.
 
+This allows you write methods more declaratively, rather than using conditionals. This feature is not only useful for recursion; you can use it to keep parallel use cases separate:
```
waterlink: condition statements?

TUTORIAL.md
```diff
@@ -434,6 +436,20 @@ end
...
+
+```ruby
+Contract lambda{|n| n < 12 } => Ticket
+def get_ticket(age)
+  ChildTicket.new(age: age)
+end
+
+Contract lambda{|n| n >= 12 } => Ticket
```
waterlink: What about using here 'catching-all-numbers' contract like `Num` ?

TUTORIAL.md
waterlink: it is just my suggestion to your comment https://github.com/egonSchiele/contracts.ruby/issues/79#issuecomment-73114220

waterlink:
Looks good to me, aside from small comments, but they are just suggestions = for you to decide.

TUTORIAL.md
kliuless: Do you mean combining contracts like `And[Num, lambda{|n| n < 12 }] => Ticket` ? I can change that. I think this is a good change, becaues it encourages people to write stricter contracts.

waterlink: Yes, that is too.

I was talking about different thing, for example, second contract could be just `Num` without lambda, because if `n < 12` fails, then it is obviously `n >= 12`, given n is a number. It is like underscore in pattern matching - last pattern catches all cases.

By the way, just to clarify, when pattern matching like this, Contracts are checked in their definition order, and method that will be executed determined by the first successful contract.

Something along these lines:

```ruby
Contract And[Num, lambda { |n| n < 12 } => Ticket
# def ...

Contract Num => Ticket
# def ...
```

On the other hand, I kinda like still having lambda in the second case, because it makes you understand when this pattern will be used without reading throughout all other patterns.

What do you think?

kliuless: I think we can use lambda in case 2, but also show `Num` as alternative, so they have a concrete example of overloads being checked in order of declaration. A simple case like this probably doesn't need a lambda for case 2, but we can mention both with a brief explanation of the benefits of each style.

waterlink: Providing these two alternatives makes sense to me!

egonSchiele:
Thanks, merged! I'm going to change `And[Num, lambda{|n| n >= 12 }]` back to `lambda{|n| n >= 12 }` if that's ok with both of you. I agree that the `And` version is technically better. But since it is a tutorial, I would like to show "the simplest case that works". So I think `And` adds unnecessary overhead that's not needed for the example.

waterlink:
@egonSchiele makes sense

egonSchiele:
cool, pushed

PR #86, opened 2015-02-09
Bug/83 :: Builtin contracts are not available on singleton classes

### LINKED ISSUE #83: Constants not available in singleton class scope
Comments speak for themselves. However, if I use the `def self.size` syntax, the error goes away.

    class TestBug
      # If removed, I get this:
      #   "In order to use contracts in singleton class, please include Contracts module in original class"
      # Ok, no problem.
      include Contracts
    
      class << self
        # If removed, I get this:
        #   in `singleton class': uninitialized constant Class::ArrayOf
        # Seems pointless to include twice.
        include Contracts
    
        Contract ArrayOf[Any] => Integer
        def size(arr)
          arr.size
        end
      end
    end
    
    p TestBug.size([1,2,3])

alex-fedorov:
Fixes #83 

During fixing this and while running a benchmark noticed that private methods defined in private section of class can't be defined because of an error, so as well added spec that exposes this and fixed it.

The fix to the original issue is as simple as including `Contracts` while `lift`ing eigenclass to contracted eigenclass.

Touched `Gemfile.lock` since it wasn't updated since last release.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    contracts (0.5)
+    contracts (0.6)
 
 GEM
   remote: http://rubygems.org/
--- a/lib/contracts/eigenclass.rb
+++ b/lib/contracts/eigenclass.rb
@@ -20,6 +20,7 @@ module Contracts
 
       unless eigenclass.respond_to?(:pop_decorators)
         eigenclass.extend(MethodDecorators)
+        eigenclass.include(Contracts)
       end
 
       eigenclass.owner_class = base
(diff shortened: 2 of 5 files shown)


filib:
@egonSchiele Any chance this could be merged and a new version pushed to `rubygems`? Unable to use `0.6` due to #89.

waterlink:
I could merge it, but I can't release to rubygems, so the only thing is to wait for @egonSchiele

filib:
@waterlink Cool, I think I'll try against the branch `bug/83/builtin_contracts_are_not_available_on_singleton_classes` in the short-term to make sure all is good.

egonSchiele:
Pushed v0.7 to rubygems!

waterlink:
:+1:

filib:
@egonSchiele Cheers!

Issue #18, opened 2013-03-22
Contracts polluting rails console


mcandre:
I'm trying to integrate a robust contract-ified gem into a new Rails app.

I added my gem as a dependency in the Rails app's `Gemspec`. I generated a scaffold for Employees. I can successfully create new employees using the Rails web interface. But when I try to use `rails console`, somehow Contracts interferes. When I remove the contract-ified gem as a dependency in the Rails app's `Gemspec`, the error goes away. And strangely, the error only presents in the `rails console` interface. Everything working okay in the Rails web interface.

Trace:

```
$ rails console
Loading development environment (Rails 3.2.13)
irb(main):001:0> Employee.create :name => "Andrew Pennebaker, :github_username => "mcandre"
RuntimeError: Couldn't find decorator for method Class:qualified_const_defined?.
Does this method look correct to you? If you are using contracts from rspec, rspec wraps classes in
it's own class.
Look at the specs for contracts.ruby as an example of how to write contracts in this case.
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/contracts-0.2.1/lib/decorators.rb:57:in `qualified_const_defined?'
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/activesupport-.2.13/lib/active_support/dependencies.rb:378:in `qualified_const_defined?'
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/activesupport-.2.13/lib/active_support/dependencies.rb:490:in `load_missing_constant'
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/activesupport-.2.13/lib/active_support/dependencies.rb:192:in `block in const_missing'
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/activesupport-.2.13/lib/active_support/dependencies.rb:190:in `each'
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/activesupport-.2.13/lib/active_support/dependencies.rb:190:in `const_missing'
from (irb):1
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/railties-.2.13/lib/rails/commands/console.rb:47:in
`start'
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/railties-.2.13/lib/rails/commands/console.rb:8:in
`start'
from c:/Ruby193/lib/ruby/gems/1.9.1/gems/railties-3.2.13/lib/rails/commands.rb:41:in `<top (required)>'
[...]

PR #87, opened 2015-02-09
Deprecation/81 :: Warning for top level inclusion of contracts

### LINKED ISSUE #81: Add deprecation warning for top-level inclusion of Contracts
- unless it is done in irb/pry

waterlink:
fixes #81 
related to #18 

This one goes not to master branch, but to `0.99-before-release`, which will be updated from master all the time, and then at some point we will want to release 0.99 with deprecations

Additionally ensures that specs pass when contracts are included in top level.


waterlink:
This is still WIP, I want to check something

waterlink:
BTW, should we support `$0 == "rails_console"` ? I still need to check out what happens to rails console when one include contracts to `Object` while in it..

lib/contracts/support.rb
```diff
@@ -1,38 +1,53 @@
...
+      def eigenclass_hierarchy_supported?
+        return false if RUBY_PLATFORM == "java" && RUBY_VERSION.to_f < 2.0
+        RUBY_VERSION.to_f > 1.8
+      end
 
+      def handle_top_level_inclusion(base)
+        return if repl?
+        return unless Object == base
+
+        STDERR.puts TOP_LEVEL_INCLUSION_DEPRECATION
```
waterlink: Any preference to use some kind of logger/presenter/formatter here instead of just STDERR ?

spec/module_spec.rb
```diff
@@ -1,5 +1,6 @@
 module Mod
   include Contracts
+  include Contracts::Modules
```
waterlink: This one didn't work with `CONTRACTS_TOP_LEVEL_INCLUSION=true`

waterlink:
@egonSchiele I think this one is ready for review.

I have used cucumber/aruba for testing out-of-process stuff, like top-level and Object inclusion, which couldn't be done in specs themselves since it will taint `Object` class for all other specs. Additionally aruba allowed for very simple interactive testing of cases with irb and pry.

egonSchiele:
Thanks! Never used Aruba before, it is pretty neat.

Issue #96, opened 2015-03-08
Tutorial missing (apparently) important detail about including in module


beezee:
I wrote a gem to define some Contracts for a few nested types, and included in the module defined by my gem.

My tests passed, but when I bundled it into a rails project rails would crash on start with the following error:

```
eigenclass.rb:13:in `lift': private method `singleton_class?' called
```

Turns out that including Contracts::Modules *before* Contracts fixes this problem, but the tutorial actually suggests the opposite order, which is what seems to have caused the error in the first place.

beezee:
Nope, confused myself and jumped the gun. I am not finished debugging this problem yet. Closing for now but if anyone recognizes the issue would love to hear.

beezee:
So it appears that activesupport defines the singleton_class? method on Class with private visibility.

It also appears that they are taking that out - https://github.com/rails/docrails/commit/dc6efd387faf7f11e93877cb6c74033d0b73e75f

But I've upgraded to rails 4.2.0 and I still have that code in my local gemset, so I'm guessing this is not released into stable on rubygems yet.

In the meantime, I'll fork and open a PR to use send from Eigenclass.lift and drop the explicit self receiver from the call in MethodDecoractors.decorate. Hopefully this is not too objectionable.

waterlink:
Fixed by #97

PR #97, opened 2015-03-08
Singleton class visibility conflict

beezee:
Addresses #96 - it defers to weird behavior in activesupport that appears to be on its way out, but activesupport is kind of a heavyweight to conflict with. Tests pass.

### DIFF
--- a/lib/contracts/decorators.rb
+++ b/lib/contracts/decorators.rb
@@ -169,7 +169,7 @@ Here's why: Suppose you have this code:
     end
 
     def decorate(klass, *args)
-      if self.singleton_class?
+      if singleton_class?
         return EigenclassWithOwner.lift(self).owner_class.decorate(klass, *args)
       end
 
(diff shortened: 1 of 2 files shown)


waterlink:
Thx, @beezee 

So `activesupport` defines its own `signleton_class?` ? And it is a private method.

In which ruby that happens? Because AFAIK in recent ruby versions this method should be already defined and public.

Other then a couple of these questions, looks good to me.

beezee:
Correct. I'm not sure why it was defined as private, as there is a native ```singleton_class``` accessor that is public, but that's life I guess.

I'm not aware of any Ruby version that defines ```singleton_class?``` natively. I found it undefined in a bare IRB session under both 1.9.3-p448 and 2.0.0-p598.

beezee:
I stand corrected, just checked the latest 2.2.1 and it is there and it is public.

Personally I'm happy to treat 2.2.1 as a requirement for contracts on rails, but the docs verify contracts on quite a few earlier versions, and a conflict with activesupport/rails seems like a pretty obvious one. I'm actually surprised this hasn't come up prior.

waterlink:
Yes, and AFAIK 2.1 has this method too. But again LGTM

PR #98, opened 2015-03-09
In the tutorial, summarize what each built-in does

roryokane:
The tutorial recommended [the RDoc](http://www.rubydoc.info/gems/contracts/Contracts) if you want to see what each built-in does, but that has the following problems:

- the contracts are mixed in with irrelevant classes like `ContractsNotIncluded` and `MethodDecorators`
- the contracts are sorted alphabetically, instead of grouping related contracts together
- each contract must be clicked on to see each description – there is no one page that shows an overview of all built-in contracts

So I wrote summaries in the tutorial, with links to the RDoc for each class.

I also changed “builtin” to “built-in”. It looks more grammatical to me.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -59,17 +59,32 @@ You can also see the contract for a function with the `functype` method:
     functype(:add)
     => "add :: Num, Num => Num"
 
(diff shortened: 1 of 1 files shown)


waterlink:
@roryokane Thanks!

LGTM

PR #99, opened 2015-03-09
Fix indentation in Contracts module documentation

roryokane:
You can currently see the problem at http://www.rubydoc.info/gems/contracts/Contracts.


waterlink:
Good catch!

PR #102, opened 2015-03-10
Add `Eq` Class for contracts around equality.

nixpulvis:
Sometimes I'd like to test an argument is equal to a defined set of values. This currently works when I'm dealing with most values, but not when my values are classes.

```ruby
class Foo
  # ...
end

Contract Foo => Any
def foo(arg)
end

foo(Foo.new)  # Works.
foo(Foo)  # Contract error.
```

This adds support for an `Eq` contract, which allows for equality comparison of classes (or anything) using `.equal?`.

```ruby
class Foo
  # ...
end

Contract Eq[Foo] => Any
def foo(arg)
end

foo(Foo.new)  # Contract error.
foo(Foo)  # Works.
```

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -3,7 +3,7 @@ require 'contracts/testable'
 =begin rdoc
 This module contains all the builtin contracts.
 If you want to use them, first:
-  
+
   import Contracts
 
 And then use these or write your own!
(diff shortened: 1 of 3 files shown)


nixpulvis:
Sorry about the whitespaces on the diff.

waterlink:
LGTM

egonSchiele:
Thank you!

PR #106, opened 2015-03-11
Add `Nat` Contract

nixpulvis:
Should really have a `Nat` contract, and should make it 100% clear that `Pos` and `Neg` do not include 0.


waterlink:
LGTM

waterlink:
Thanks @nixpulvis !

PR #107, opened 2015-03-13
Improved arg validation parsing.

### LINKED ISSUE #90: Pattern matching with different arity fails
```ruby
Contract Num, String => String
def do_stuff(number, string)
  # .. do stuff ..
end

Contract Num, String, Num => String
def do_stuff(number, string, other_number)
  # .. do different stuff ..
end

do_stuff(3, "abc", 4)    #=> ArgumentError: wrong number of arguments (3 for 2)
```

Happens, because first pattern matching contract does not fail. And it probably should fail by number of arguments.

### LINKED ISSUE #100: Splat with named argument fails.
Hi, I came across this years ago and a lecturer just reminded me of it. Finally I have a project that could benefit from Contracts. Really impressive work.

I've got some complex method signatures with splats, named arguments and doublesplats. I can't get Contracts to work for certain combinations and have distilled to a simple example:

```ruby
Contract Args[Num], { operation: Symbol } => Num
=> [[Contract, [#<Contracts::Args:0x007ffa24adfe78 @contract=Contracts::Num>, {{:operation=>Symbol}=>Contracts::Num}]]]
def nums(*nums, operation: :+)
  nums.inject(&operation)
end
=> :nums
nums(1,3,5,6)
ContractError: Contract violation for argument 2 of 4:
    Expected: {:operation=>Symbol},
    Actual: 3
    Value guarded in: Object::nums
    With Contract: Args, Hash => Num
```

Am I using Contracts wrong or is this a bug? 

A more complex signature from my actual project that hopefully fixing the above will solve:

```ruby
Contract Args[Or[Array, N]], { closed: Maybe[Bool] } => Path
def initialize(*nodes, closed: true, **options)
etc...
```

Thanks

sfcgeorge:
Complex method signatures that Ruby allows should now work with Contracts, which fixes #100. Pattern matching with differing arities now works too, which fixes #90. Also fixed some tests that didn't test what they meant to, and added separate test files for syntax only supported by newer Rubies.

## Complex Method Signatures

Before, Contracts made some assumptions about method signatures that were stricter than what Ruby allows. This meant that many methods could not be annotated with Contracts. The problem was assuming that a splat always comes last, when it need not.

I fixed this by looping forwards validating the args up to the splat (or to the end if there isn't one), then looping backwards validating to the splat, then fixing the validator to the splat one and validating any remaining args with that.

This worked, all existing tests pass plus the handful I added. Complex methods such as this now work:

```ruby
  Contract String, Bool, Args[Symbol], Float, { e:Range, f:Maybe[Num] }, Proc =>
           [Proc, Hash, Num, Range, Float, ArrayOf[Symbol], Bool, String]
  def complicated(a, b=true, *c, d, e:, f:2, **g, &h)
    h.call [h, g, f, e, d, c, b, a]
  end
```

## Pattern Matching Different Arities

```ruby
Contract Num, String => String
def do_stuff(number, string)
  "foo"
end

Contract Num, String, Num => String
def do_stuff(number, string, other_number)
  "bar"
end

do_stuff(3, "abc", 4)
=> "bar"
do_stuff(5, "def")
=> "foo"
```

## Speed comparison

Master:

```
testing add                      0.310000   0.000000   0.310000 (  0.305342)
testing contracts add            2.950000   0.000000   2.950000 (  2.954473)

#call_with         | 0.027 ms | 0.348 ms | 0.035 ms     | 345.975 ms
```

My modifications:

```
testing add                      0.310000   0.000000   0.310000 (  0.311739)
testing contracts add            2.810000   0.000000   2.810000 (  2.812024)

#call_with         | 0.047 ms | 0.478 ms | 0.061 ms     | 607.779 ms
```

So my changes to the `call_with` method seem to have made it almost twice as slow, but overall it has had no noticeable effect on performance.

## Improved Tests

In trying to work out what a 16 line chunk of code did I commented it out and all the tests still passed. However after some manual testing in IRB I figured out what it was for, so I added tests for this before rewriting the arg parsing code. The problem before was having multiple arg validations in one test, so when a test was expecting to raise an exception, it could have been any of the args. Making just the arg validation type being tested a strict validation and the rest `Any` helps fix this. It's also a good idea to change types between args and return to make sure things aren't getting muddled up, e.g. test `Num => String` rather than `Num => Num`.

## Ruby Support

I have tested my modifications all the way back to Ruby 1.8.7 and it still works! The bad news is that Rubies since 1.8.7 introduced new backwards incompatible arg syntax so tests for this new functionality can't be run on old Rubies. 1.9 added the ability for splats to come in places other than the end. 2.0 added required named arguments and doublesplats, 2.1 added optional named arguments. So I have created separate test files for these Ruby version specific tests and they will be run only on supported Rubies according to `RUBY_VERSION`. Travis is happy so it seems to have worked.

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -83,10 +83,18 @@ class Contract < Contracts::Decorator
     else
       fail "It looks like your contract for #{method} doesn't have a return value. A contract should be written as `Contract arg1, arg2 => return_value`."
     end
-    @klass, @method= klass, method
-    @has_func_contracts = args_contracts.index do |contract|
-      contract.is_a? Contracts::Func
+    @klass, @method = klass, method
+    @splat_lower_index = @args_contracts.index do |contract|
+      contract.is_a? Contracts::Args
     end
+    last_contract = @args_contracts.last
+    @has_proc_contract = Contracts::Func === last_contract || (
+        Class === last_contract &&
+        (last_contract <= Proc || last_contract <= Method)
+      )
+    penultimate_contract = @args_contracts[-2]
+    @has_options_contract = @has_proc_contract ? Hash === penultimate_contract :
+                                                 Hash === last_contract
   end
 
   def pretty_contract c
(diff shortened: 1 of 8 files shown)


sfcgeorge:
Actually when I ran the tests on 2.0 I must have left the new syntax ones commented out as it's failing on Travis. Required keyword args didn't come until Ruby 2.1, so that test should either be commented out or modified to have a default value then it will pass. 

As JRuby also fails (I hadn't thought that JRuby is a bit behind) maybe it would be best to just comment out the offending new syntax tests. It's a shame but they can be uncommented if the `call_with` method ever needs to be modified again to make sure there are no regressions in modern Rubies. 

Let me know if you want me to modify this pull request in any way, such as commenting out the new syntax tests.

egonSchiele:
Woot, thank you! I will check this out soon.

sfcgeorge:
Thank you :D I was working on something a little different to Contracts, but it grew similar due to Ruby limitations so I decided to just use Contracts, it's beautifully simple. I've added it throughout another project replacing much longer validations. 

In adding my modified branch of Contracts to my project I found 1 more bug that I didn't address with this pull request. The bug is named arguments weren't allowed to be optional, an empty hash needed inserting at the end of the args like we do for nil blocks. I have since added a test and fixed that too, ready for another pull request. 

I'm thinking of closing this pull, rebasing if possible and submitting a new pull request consisting of this one + the extra bug fix. While I'm at it, should I do something about the tests failing in old Ruby?

My idea is putting the failing tests in eval strings (nasty I know but that's the only way to stop the syntax errors on parse) then have a conditional on RUBY_VERSION to only do those tests on versions of Ruby that support those syntax features. I wondered if Rspec or Travis have a built in way of running certain tests on certain Rubies but I couldn't find anything. Conditionally testing dependent on available features has got to be better than completely commenting out the tests.

waterlink:
You don't need to put them in strings, just put them in separate files (probably in separate folder) and then apply different default RSpec file filter for different ruby versions. So that it will not even try to load these files for old ruby versions.

sfcgeorge:
@waterlink Perfect! It looks like currently there's a require glob in `spec_helper.rb` so I'll conditionally remove the unsupported files from that. Best of both worlds, full Ruby support and good test coverage.

sfcgeorge:
Ok I committed the fix for the bug I found and Contracts is now working great in a real world project of mine. Tests added and tests made to run only on Rubies that support the syntax. Rebased the commits, that seems to have automatically updated this pull request so Travis is passing and hopefully this is ready to merge.

waterlink:
I like this so far. I still can perfectly understand `call_with` (even so we will have to do something about this method in the future).

Solution to specific ruby tests is elegant.

From me it is :+1:

@egonSchiele can you take a look at this one?

waterlink:
BTW, I have a feeling, that this will eliminate most of the problems for #90

sfcgeorge:
@waterlink Thank you, I tried to only change what was necessary in new `call_with` and stick to a similar style. Refactoring would be nice but the performance blog post suggests that it has to be kept inline for speed.

Aha! It *almost* fixed #90 but I wasn't checking that the contract for a particular arg actually exists before calling it so was getting "NoMethodError: undefined method `[]' for nil:NilClass". However I just fixed it, really easy just added a nil check in 2 places and now it works! Give me half an hour to write tests for #90 and push again. Teaser:

```ruby
irb(main):001:0> include Contracts
=> Object
irb(main):002:0>
irb(main):003:0* Contract Num, String => String
=> [[Contract, [Contracts::Num, {String=>String}]]]
irb(main):004:0> def do_stuff(number, string)
irb(main):005:1>   "foo"
irb(main):006:1> end
=> :do_stuff
irb(main):007:0>
irb(main):008:0* Contract Num, String, Num => String
=> [[Contract, [Contracts::Num, String, {Contracts::Num=>String}]]]
irb(main):009:0> def do_stuff(number, string, other_number)
irb(main):010:1>   "bar"
irb(main):011:1> end
=> :do_stuff
irb(main):012:0>
irb(main):013:0* do_stuff(3, "abc", 4)
=> "bar"
irb(main):014:0> do_stuff(5, "def")
=> "foo"
```

sfcgeorge:
And there we are, tests and fix for #90 are now included too.

alex-fedorov:
Nice!

--
Best Regards,
Alexey Fedorov,
Sr Ruby Developer, Clojure Developer, Rust Developer,
+49 15757 486 476

2015-03-14 20:35 GMT+02:00 Simon George <notifications@github.com>:

filib:
:+1:

lib/contracts.rb
```diff
@@ -241,62 +249,92 @@ def call(*args, &blk)
...
+            :contract => contract,
+            :class => @klass,
+            :method => @method,
+            :contracts => self,
+            :arg_pos => args_size - i,
+            :total_args => args_size
+          })
+        end
+
+        if contract.is_a?(Contracts::Func)
```
egonSchiele: Some code duplication here

egonSchiele:
1. this is a great contribution that fixes a lot of things! Thanks.
2. good point about tests: I should avoid `Num => Num` contracts.
3. `call_with` is a jungle. As you noticed I refactored it for speed, but maybe I can strike a better balance for readability.
4. I added a note about code duplication, I'll just add a TODO for this.

Nice PR overall, merging.

sfcgeorge:
@egonSchiele Fantastic, thank you for merging!
3. Choosing between readability and speed is tough, it's got slower since your blog post as features have been added. It might be tough to speed up, simplify and refactor but I guess that's the ideal.
4. I agree. There are 2 very similar loops, but subtle differences make them tricky to generalise. I tried making a lambda for them but that broke the return statement so I left it as is.

PR #109, opened 2015-03-16
Improve to_s formatting in error's Expected field.

### LINKED ISSUE #108: to_s isn't called on contracts within Hash / Array
Currently `inspect` is called on any Hash contracts and nothing more, so any nice custom `to_s` descriptions of the nested contracts aren't called. 

```ruby
{ binary: Or[1, 0] } 
# => Expected: {:binary=>#<Contracts::Or:0x007fc4639abee0 @vals=[1, 0]>}
```

I would like nicer Expected messages, so `to_s` to be called on each Class contract within any Hash contracts. Something like this:

```ruby
{ binary: Or[1, 0] }.map{ |k, v| [k, v.class==Class ? v.to_s : v] }.to_h
# => Expected: {:binary=>"1 or 0"}
```

It would also need to be recursive as there may be Hashes within Hashes, and the same improvement should probably be done for Array contracts too.

The current code is in `contracts.rb:114`:

```ruby
   expected = if data[:contract].to_s == "" || data[:contract].is_a?(Hash)
                data[:contract].inspect
              else
                data[:contract].to_s
              end
```

Perhaps by doing this it's unclear whether the string is a description or the actually expected string, e.g. in the example above is "1 or 0" telling us we need 1 or 0 or that we need the string "1 or 0". The above dummy solution doesn't work as it only works for class contracts—you want to call `to_s` on `Or[]` and such instance contracts but not on instances of primitives like 42 otherwise it will look like a string. I'm not sure if it's possible to solve these cleanly. Thoughts?

sfcgeorge:
Fixes #108. `to_s` is now called on contracts within Hashes and Arrays.

to_s is recursively called on Hashes and Arrays, so even nested Hashes or Arrays will contain pretty contract representations.

Contracts module prefix is now removed from classes, this makes more sense to me as when you write a contract such as `Contract Num => nil` you'd expect the error to be `Expected: Num` rather than `Expected: Contracts::Num`.

Custom to_s messages will be wrapped in round brackets to differentiate from standard Strings. E.g. `Expected: (a number)`. Without this custom messages get confusing in nested hashes or arrays. Alternatives might be angle brackets (but these get confused with hash rocket syntax) or backticks.

Primitive values e.g. 42, true, nil will be left alone, no erroneous conversion to string.

Tests added for all of these cases. Rebased onto current master. Let me know if you want the coding style changing or want to discus any of the decisions I made.

### DIFF
--- a/lib/contracts/formatters.rb
+++ b/lib/contracts/formatters.rb
@@ -1,0 +1,69 @@
+module Contracts
+  # A namespace for classes related to formatting.
+  module Formatters
+    # Used to format contracts for the `Expected:` field of error output.
+    class Expected
+      def initialize(contract)
+        @contract = contract
+      end
+
+      # Formats any type of Contract.
+      def contract(contract = @contract)
+        if contract.is_a?(Hash)
+          hash_contract(contract)
+        elsif contract.is_a?(Array)
+          array_contract(contract)
+        else
+          ContractInspectWrapper.new(contract)
+        end
+      end
+
+      # Formats Hash contracts.
+      def hash_contract(hash)
+        hash.inject({}) { |repr, (k, v)|
+          repr.merge(k => ContractInspectWrapper.new(contract(v)))
+        }.inspect
+      end
+
+      # Formats Array contracts.
+      def array_contract(array)
+        array.map{ |v| ContractInspectWrapper.new(contract(v)) }.inspect
+      end
+    end
+
+    # A wrapper class to produce correct inspect behaviour for different
+    # contract values - constants, Class contracts, instance contracts etc.
+    class ContractInspectWrapper
+      def initialize(value)
+        @value = value
+      end
+
+      # Inspect different types of contract values.
+      # Contracts module prefix will be removed from classes.
+      # Custom to_s messages will be wrapped in round brackets to differentiate
+      # from standard Strings.
+      # Primitive values e.g. 42, true, nil will be left alone.
+      def inspect
+        return @value.to_s if plain?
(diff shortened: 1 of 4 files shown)


lib/contracts/formatters.rb
```diff
@@ -0,0 +1,65 @@
...
+        @value = value
+      end
+
+      # Inspect different types of contract values.
+      # Contracts module prefix will be removed from classes.
+      # Custom to_s messages will be wrapped in round brackets to differentiate
+      # from standard Strings.
+      # Primitive values e.g. 42, true, nil will be left alone.
+      def inspect
+        if @value.is_a?(CallableClass) || @value.class == Class
```
alex-fedorov: I would rewrite it with guards:

```ruby
def inspect
  return @value.to_s if plain?
  return "(#{@value.to_s})" if has_useless_to_s?
  @value.inspect.gsub(/^Contracts::/, '')
end

private

def plain?
  !@value.is_a?(CallableClass) && @value.class != Class
end

def has_useless_to_s?
  @value.to_s == "" || @value.to_s == @value.inspect
end
```

alex-fedorov: Updated example, useless => useful

alex-fedorov:
That looks very nice! Left one comment, but not a merge blocker.

sfcgeorge:
@alex-fedorov I like that refactor so I included it, much cleaner. Not sure what you meant by Updated example, useless => useful

alex-fedorov:
I have edited code in previous code - that what I meant

egonSchiele:
Looks good to me. Thank you!

PR #111, opened 2015-03-17
Adit/refactoring

egonSchiele:
I'm adding this PR to spur discussion. To me this code is more readable. But there is a real performance hit. Old, optimized code:

    testing add                      0.370000   0.000000   0.370000 (  0.371734)
    testing contracts add            4.250000   0.000000   4.250000 (  4.249101)

New, readable code:

    testing add                      0.360000   0.000000   0.360000 (  0.364561)
    testing contracts add            9.000000   0.020000   9.020000 (  9.014563)

It is twice as slow on my machine, and much much slower than `add` without contracts. Should I optimize for readability or speed here? WDYT @waterlink ?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,5 +10,3 @@ group :development do
   gem "method_profiler"
   gem "ruby-prof"
 end
-
-gem "memoizable"
--- a/Gemfile
+++ b/Gemfile
@@ -10,3 +10,5 @@ group :development do
   gem "method_profiler"
   gem "ruby-prof"
 end
+
+gem "memoizable"
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -364,7 +364,12 @@ class Contract < Contracts::Decorator
     end
 
     unless ret_validator[result]
-      Contract.failure_callback({:arg => result, :contract => ret_contract, :class => klass, :method => method, :contracts => self, :return_value => true})
+      Contract.failure_callback({:arg => result,
+                                 :contract => ret_contract,
+                                 :class => klass,
+                                 :method => method,
+                                 :contracts => self,
+                                 :return_value => true})
     end
 
     this.verify_invariants!(method) if this.respond_to?(:verify_invariants!)
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -8,6 +8,7 @@ require 'contracts/invariants'
 require 'contracts/method_reference'
 require 'contracts/modules'
 require 'contracts/support'
+require 'memoizable'
 
 module Contracts
   def self.included(base)
@@ -63,6 +64,7 @@ end
 #
 # This class also provides useful callbacks and a validation method.
 class Contract < Contracts::Decorator
+  include ::Memoizable
   # Default implementation of failure_callback. Provided as a block to be able
   # to monkey patch #failure_callback only temporary and then switch it back.
   # First important usage - for specs.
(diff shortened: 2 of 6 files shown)


egonSchiele:
Also pinging @sfcgeorge since you just touched this code...any strong feelings?

lib/contracts/builtin_contracts.rb
```diff
@@ -122,7 +124,7 @@ def valid?(val)
     end
 
     def to_s
-      @vals[0, @vals.size-1].join(", ") + " or " + @vals[-1].to_s
+      @vals[0, @vals.size-1].map { |x| InspectWrapper.new(x) }.join(", ") + " or " + InspectWrapper.new(@vals[-1]).to_s
```
sfcgeorge: Nice idea. It's turning out to be a handy class, I may use it in YARD output too.

lib/contracts/formatters.rb
```diff
@@ -44,6 +44,7 @@ def initialize(value)
       # from standard Strings.
       # Primitive values e.g. 42, true, nil will be left alone.
       def inspect
+        return @value.inspect if empty_val?
```
sfcgeorge: I'm not sure this is needed. Please add a test case showing that the formatting would look wrong without it, see other tests I added.

Gemfile
```diff
@@ -10,3 +10,5 @@ group :development do
   gem "method_profiler"
   gem "ruby-prof"
 end
+
+gem "memoizable"
```
sfcgeorge: I'm unsure about adding a dependency, it doesn't seem necessary. I'm not too keen on the syntax of Memoizable, I feel like `||=` would be just as good and probably perform slightly faster, e.g.

```ruby
def ret_validator
  @ret_validator ||= Contract.make_validator(ret_contract)
end
```

Factoring out these chunks as methods has improved the code though, so whether or not Memoizable is used to do it probably doesn't matter, I'm just edgy about it. On the plus side it looks well maintained and supports the same Rubies we do.

lib/contracts.rb
```diff
@@ -242,108 +267,110 @@ def call(*args, &blk)
...
-      splat_upper_index = @splat_lower_index
-      (_args_size - @splat_lower_index).times do |i|
-        _arg = _args[_args_size - 1 - i]
+    if args_contract_index
+      splat_upper_index = args_contract_index
+      (args.size - args_contract_index).times do |i|
+        arg = args[args.size - 1 - i]
 
-        if Contracts::Args === @args_contracts[contracts_size - 1 - i]
+        if Contracts::Args === args_contracts[args_contracts.size - 1 - i]
```
sfcgeorge: It might be worth benchmarking a cached value of size vs calculating it each time as the value is used a few times. Ruby may internally optimize it or it may be negligible.

lib/contracts.rb
```diff
@@ -242,108 +267,110 @@ def call(*args, &blk)
...
+
+    # Explicitly append options={} if Hash contract is present
+    maybe_append_options!(args, blk)
 
+    # Loop forward validating the arguments up to the splat (if there is one)
+    (args_contract_index || args.size).times do |i|
+      contract = args_contracts[i]
+      arg = args[i]
+      validator = args_validators[i]
+      return unless validate_arg(arg, contract, validator, i+1, args)
```
sfcgeorge: This reads much better, but as your blog post shows method calls are slow, calling `validate_arg` in a loop may be unacceptable. Benchmark, inline it then bench again to see if it's a bottle neck, that's my hunch. I wonder if a lambda local variable would be faster than a full method here as an alternative.

sfcgeorge:
Over all I like your changes, you've made some sensible refactors and simplified things I wondered if could be simplified. I made a few picky code comments but it's a net positive from me, reading through it was easier to make sense of. Really nice.

The slow down is worrying though. Over 2x slower is a big hit. My feeling is that this is a low level tool where it should simply be possible to add it to an existing project with minimal impact (on anything; coding style, structure, speed etc).

Perhaps re-optimize from here—certainly a number of your changes can stay—changing the just the slowest bottlenecks for less readable but faster alternatives. At least it's mainly just this method that needs to be optimised; other parts of the library can be coded beautifully without worry for speed as they'll only be run a handful of times. `call_with` gets run a lot.

Example; The project I've added Contracts to is a design by code tool, users code `square`, `circle` etc and it gets rendered. The smallest primitive is a Node which is a light class that just wraps x, y coordinates and type (for corners, curves etc). As this is a user facing tool I use Contracts to validate the constructor for Nod, and a canvas can have thousands of nodes. Rendering large scenes can take 5 seconds. It's an extreme but real world example that sometimes Contracts will be added to simple methods that get called a lot, and ideally it would still be fast.

waterlink:
This is huge improvement over readability, I would like to add a benchmark
with some simple IO. Like writing and reading from file, or doing something
with redis, or just fetching google page over the wire.

Have you given a thought about code generating techniques? So that we can
code nice pseudo-class with bunch of small methods, but then they all will
get inlined in one huge effective method.

alex-fedorov:
BTW I have strong feelings, that gem memoizable is not needed at all, it is
a runtime dependency people will not want and it can be replaced with
simple `||=` or `if defined?` for nil, false aware memoizing.

sfcgeorge:
Code generation is an interesting idea. I was working on something similar until I found contracts and it does work that way, `class_eval` a method definition string with all the validations done as inline if statements—that way the result is exactly what you'd write by hand so as fast as possible. Generating that kind of code is nasty though, and I don't think it would work with our dynamic method dispatch. Nice idea for speed, but probably not practical to implement.

waterlink:
Not generating code on runtime, but actually generating Ruby file from
pseudo-ruby file.

sfcgeorge:
Oh I see your point, that could be pretty nice. Make the code readable, annotate methods to be inlined and run it through a tool. Some languages can do inlining with an annotation or automatically (Rubinius aparently), alas I think MRI can't. I feel like there should be a gem to do this high level Ruby => performant Ruby inlining but I can't find such a thing.

alex-fedorov:
I would just fix all the minor code review comments in this PR and merge it. Then as a next step we can think about looking for this tool. If there is no such thing out there, we can build it. Sounds like a plan.

Gemfile
nixpulvis: I'd also strongly advise against adding dependencies, unless they truly are needed. This doesn't seem like a case where it's worth it.

egonSchiele: noted and removed.

waterlink:
Something is strange with travis, it just doesn't want to run build..

sfcgeorge:
It did that for me, probably just a busy time.

waterlink:
LGTM from me. I think this can be merged when green.

egonSchiele:
Original time for refactored code:

    testing add                      0.380000   0.000000   0.380000 (  0.382351)
    testing contracts add            7.680000   0.010000   7.690000 (  7.702954)

I changed the memoized methods back to being instance variables again, and got a big boost:

    testing add                      0.380000   0.000000   0.380000 (  0.374630)
    testing contracts add            4.650000   0.010000   4.660000 (  4.669573)

This refactoring still adds improvements, so I think I will merge it as-is. Like @sfcgeorge said this is a low-level library so I want it to have as little performance impact as possible. I like the idea of code generation for the future! We will just have to be careful so the code doesn't end up more complicated than before.

I also added IO benchmarks (downloading a web page 100 times). As expected, contracts makes no noticeable difference there:

    testing download                 2.090000   0.180000   2.270000 ( 33.597591)
    testing contracts download       2.180000   0.160000   2.340000 ( 30.947244)

The goal is to be fast enough that people don't stop using contracts for performance reasons.

egonSchiele:
merged.

sfcgeorge:
Great! It's fantastic that removing Memoize gave such a speed boost, and slightly surprising as it's supposed to speed things up but I guess it abstracts too much. This refactor is barely slower than before, 11.5x VS 12.5x, which is good news.

Code generation still sounds interesting to me though it could add more complexity than it's worth. Definitely one to have a think about and experiment.

PR #114, opened 2015-03-23
Add initial YARD support

### LINKED ISSUE #32: Add YARD support
While I love writing contracts in my code, I find that I must duplicate my type annotations when documenting that same code. I use [YARD](http://yardoc.org/) for generating documentation which, allegedly, [supports special DSLs in Ruby](http://yardoc.org/guides/extending-yard/writing-handlers.html).

It would really rock if contracts.ruby shipped with YARD support to make documentation easier! Happy to help beta test this feature and contribute however I can.

Thanks!

sfcgeorge:
YARD documentation generation for Contracts has been made a gem [yard-contracts](https://github.com/sfcgeorge/yard-contracts) Fixes #32 

This pull request adds a minor feature to the Formatters. The new `full` parameter means inspect will output an empty string (when `full=false`) if the `to_s` value would be useless. This is to prevent duplication in yard-contracts; A param docstring is the type (inspect) followed by the description (to_s). Without this patch the type and description can end up being the same which looks odd.

In MRI Ruby < 2 inspect calls to_s so they _always_ are the same. I couldn't find a way to make this work in those old Rubies. JRuby 1.9mode is fine. It doesn't affect the main Contracts gem because when they are the same it calls inspect, but as inspect calls to_s it happens to work as expected.

Currently I have duplicated/overridden this code in yard-contracts so it works for now without this pull request. If you decide to merge then I can take this bit out of yard-contracts.

### DIFF
--- a/lib/contracts/formatters.rb
+++ b/lib/contracts/formatters.rb
@@ -5,7 +5,7 @@ module Contracts
     class Expected
       # @param full [Boolean] if false only unique `to_s` values will be output,
       #   non unique values become empty string.
-      def initialize(contract, full=true)
+      def initialize(contract, full = true)
         @contract, @full = contract, full
       end
 
@@ -23,15 +23,15 @@ module Contracts
       # Formats Hash contracts.
       def hash_contract(hash)
         @full = true # Complex values output completely, overriding @full
-        hash.inject({}) { |repr, (k, v)|
+        hash.inject({}) do |repr, (k, v)|
           repr.merge(k => InspectWrapper.new(contract(v), @full))
-        }.inspect
+        end.inspect
       end
 
       # Formats Array contracts.
       def array_contract(array)
         @full = true
-        array.map{ |v| InspectWrapper.new(contract(v), @full) }.inspect
+        array.map { |v| InspectWrapper.new(contract(v), @full) }.inspect
       end
     end
 
@@ -40,7 +40,7 @@ module Contracts
     class InspectWrapper
       # @param full [Boolean] if false only unique `to_s` values will be output,
       #   non unique values become empty string.
-      def initialize(value, full=true)
+      def initialize(value, full = true)
         @value, @full = value, full
       end
 
@@ -53,7 +53,7 @@ module Contracts
         return '' unless full?
         return @value.inspect if empty_val?
         return @value.to_s if plain?
-        return delim(@value.to_s) if has_useful_to_s?
+        return delim(@value.to_s) if useful_to_s?
         @value.inspect.gsub(/^Contracts::/, '')
       end
 
@@ -67,14 +67,15 @@ module Contracts
       end
 
       private
+
       def empty_val?
-        @value.nil? || @value == ""
+        @value.nil? || @value == ''
       end
 
       def full?
         @full ||
-        @value.is_a?(Hash) || @value.is_a?(Array) ||
-        (!plain? && has_useful_to_s?)
+          @value.is_a?(Hash) || @value.is_a?(Array) ||
+          (!plain? && useful_to_s?)
       end
 
       def plain?
@@ -82,10 +83,10 @@ module Contracts
         !@value.is_a?(CallableClass) && @value.class != Class
       end
 
-      def has_useful_to_s?
+      def useful_to_s?
         # Useless to_s value or no custom to_s behavious defined
         # Ruby < 2.0 makes inspect call to_s so this won't work
-        @value.to_s != "" && @value.to_s != @value.inspect
+        @value.to_s != '' && @value.to_s != @value.inspect
       end
     end
   end
--- a/lib/contracts/formatters.rb
+++ b/lib/contracts/formatters.rb
@@ -3,8 +3,10 @@ module Contracts
   module Formatters
     # Used to format contracts for the `Expected:` field of error output.
     class Expected
-      def initialize(contract)
-        @contract = contract
+      # @param full [Boolean] if false only unique `to_s` values will be output,
+      #   non unique values become empty string.
+      def initialize(contract, full=true)
+        @contract, @full = contract, full
       end
 
       # Formats any type of Contract.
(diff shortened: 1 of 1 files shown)


vendethiel:
This really looks good! :cake:

alex-fedorov:
Hi @sfcgeorge. That looks awesome! :+1: 

I would vote for this to be extracted as a `yard-contracts` gem. Here you can add any modifications to `InspectWrapper` and then mark it and methods you care about with `# @public`, which will make it subject to semantic versioning (once 1.0.0 is out, we couldn't break this (we could, but this is subject to fixing in next patch version as soon as possible)).

I would even volunteer to maintain `yard-contracts` :)

If you extract it as a different gem, then lack of the tests is not a big problem, we will be able to do that once the thing is going - probably with `aruba` and `cucumber`.

About ruby 2.2, this comment says that bug is fixed in 2.2.1: https://github.com/lsegal/yard/issues/825#issuecomment-83117467

I would leave here changes to `InspectWrapper` and put all YARD and AST-parsing logic to the new gem.

Everything else can be figured out by iterating over it. Just having this core thing is awesome.

sfcgeorge:
@alex-fedorov Fantastic, glad you like.

Okay I agree a separate gem is best. Tests can be written with whatever tool. It can be expanded without overcomplicating Contracts. `yard-contracts` could be put in a bundler group so it isn't even loaded in production. It can have a separate README and docs for clarity.

It would be great to work with you on it. The remaining question is where. I'm happy to package the gem and could certainly put it under my account, adding @alex-fedorov and @egonSchiele as maintainers (if you want). Or perhaps @egonSchiele would rather have it under their account. Or there's the option of a Contracts organisation to group contracts related projects, I don't know exactly how that works.

It's just an admin issue, I don't mind either way. For now I guess I can put the gem under my account, then if we decide to put it elsewhere it can be forked with a redirect.

alex-fedorov:
Yeah, that works, we can figure it out later, when we have opinion from @egonSchiele.

egonSchiele:
This is a really great first stab! I tested it out and it's great to see contract info in the docs...makes them really useful. Another issue I found:

Right now it doesn't work with custom contracts. For example, if I create the `Rootable` contract, its `to_s` message won't get included in the docs. That's because of the way `get_contract_value` works. It gets a string, and tries to check if a constant with that name exists. But when yardoc is generating the docs for a file, that file hasn't been loaded. So the `Rootable` class won't be loaded :-/

So it only works with built-in contracts, since those get loaded when we require yard_extensions.rb. One solution is to have users put all their contracts in a separate file, and then that file could be loaded when generating docs:

```ruby
yardoc -e ../contracts.ruby/lib/yard_extensions.rb -e types.rb
```

Now if I put `Rootable` in `types.rb`, it will get loaded and evaluated correctly. But there's a new issue: what if `Rootable` is in a module?

```ruby
module Types
  class Rootable
    ...etc...
  end
end
```

And I use it like this:

```ruby
class Test
  include Contracts
  include Types

  # Find the root of a number, like `Math.sqrt` for arbitrary roots.
  Contract Pos, Rootable => Rootable
  def root(root, num)
    num**(1.0/root)
  end
end
```

`get_contract_value` won't be able to find the constant `Rootable` because it is actually `Types::Rootable`.

Overall I think this is extremely useful. I'm with @waterlink, I suggest making this a new gem, since it is complex enough that it will have its own set of features, tests, and issues. It could be released as-is as `v0.1`, and then we could work through the list of TODOs you posted.
I would rather have you or @waterlink as owners/maintainers, but I'm happy to contribute to it.

sfcgeorge:
Hi @egonSchiele thanks for testing it out!

Yep I had thought about that issue with loading custom contract classes and came to the same conclusion as you but forgot to mention it. Putting all your custom contracts in a single file (probably a good idea anyway) and giving it to YARD with the -e flag works nicely. If you don't want them under the global namespace you can put them under the Contracts namespace and it works fine.

Ok, I've wrapped it up as a gem that can now be used with YARD simply by adding the flag `--plugin contracts`, nice and simple. I haven't pushed to Rubygems yet, I'll probably wait until I've added tests. 

I'll amend this pull request to update the `InspectWrapper` but for now I copied the Formatters into the new gem so that it works right away. 

I've added @egonSchiele and @waterlink as collaborators just in case you want to help out.

egonSchiele:
Awesome :+1:

sfcgeorge:
Updated pull request to just include what is useful to yard-contracts now it has been extracted.

lib/contracts/formatters.rb
```diff
@@ -60,6 +71,12 @@ def empty_val?
         @value.nil? || @value == ""
       end
 
+      def full?
+        @full ||
+        @value.is_a?(Hash) || @value.is_a?(Array) ||
```
waterlink: Wrong indentation here, should be:

```ruby
condition_a ||
  condition_b ||
  condition_c ||
  # ...
```

waterlink:
Looks good to me

sfcgeorge:
@waterlink I couldn't find anything about that in ruby-style-guide, is that your preference?

```ruby
#1
@full ||
@value.is_a?(Hash) || @value.is_a?(Array) ||
(!plain? && has_useful_to_s?)

#2
@full ||
  @value.is_a?(Hash) || @value.is_a?(Array) ||
  (!plain? && has_useful_to_s?)

#3
@full ||
  @value.is_a?(Hash) || @value.is_a?(Array) ||
    (!plain? && has_useful_to_s?)

#4
@full ||
  @value.is_a?(Hash) ||
  @value.is_a?(Array) ||
  (!plain? && 
  has_useful_to_s?)
```

I prefer style 1 above, but I don't mind changing it though to be consistent with you. I assume it's style 2 you'd like me to make it?

waterlink:
Style 2 is good, if you want, we can wait for @egonSchiele to decide.

Style 1 is clearly bad because at the first glance it looks like 3 different expressions

waterlink:
Hm, strange I can't find this rule in popular ruby style guides, but that is actually a first time I see anybody using this kind of chaining of `||` in ruby for a long time..

nixpulvis:
I think 1 is the best because it most directly follows other rules about aligning elements. There is no reason that expressions that come after the first expression should be indentet.

sfcgeorge:
Oh dear I've started something ;) I do this kind of thing quite a bit, maybe it's a sign it should be refactored, I don't know. Usually this kind of chaining happens in an `if` statement in which case I do like to indent it.

nixpulvis:
There is nothing wrong with a `||` chain here and there. But in this case I might agree you could make that function make a bit more sense. I wouldn't overthink it though.

There is no reason an `if` would be any better. In fact if you are returning a Boolean I'd argue it only serves to make things harder to understand.

Indentation to make clear that it's all one chain doesn't seem bad to me though, I guess I just see this issue in two lights. Might be worth opening an issue over at https://github.com/bbatsov/ruby-style-guide

sfcgeorge:
Okay! Something more quantitative. I installed [Rubocop](https://github.com/bbatsov/rubocop) and got it working with [Syntastic](https://github.com/scrooloose/syntastic) in Vim. Rubocop is a static code analyser based on the ruby-style-guide. It suggested a number of style improvements which I have made---including @waterlink's assertion that multiline statements should be indented. 

I suggest we aim to follow the ruby-style-guide / Rubocop as so far it seems sensible and I agree with it. There is an open issue to use Reek for style checking, that seems less popular, there are others too.

egonSchiele:
Looks fine to me, but what is the `full?` method used for checking? Is it only used in yard-contracts?

sfcgeorge:
Yep, only yard-contracts. So that in the output you get:

```ruby
(Num)
(Or[String, Symbol]) - String or Symbol
```

Instead of:

```ruby
(Num) - Num
(Or[String, Symbol]) - String or Symbol
```

With the duplicated `Num` as it has no custom to_s. Do you think it shouldn't be in Contracts if it isn't used directly?

egonSchiele:
I'll just add a comment. Thank you!

Issue #21, opened 2013-05-29
reek warnings


mcandre:
I took a look with the popular Ruby linter [reek](https://github.com/troessner/reek) to examine potential sources of bugs. There actually aren't that many :)

    $ reek -q lib
    lib/builtin_contracts.rb -- 11 warnings:
    Contracts::::Hash has no descriptive comment (IrresponsibleModule)
    Contracts::::Hash#test_data contains iterators nested 2 deep (NestedIterators)
    Contracts::::Hash#test_data has approx 12 statements (TooManyStatements)
    Contracts::::String has no descriptive comment (IrresponsibleModule)
    Contracts::Any#self.valid? has unused parameter 'val' (UnusedParameters)
    Contracts::Args#test_data calls Testable.test_data(@contract) 3 times (DuplicateMethodCall)
    Contracts::ArrayOf#test_data calls Testable.test_data(@contract) 3 times (DuplicateMethodCall)
    Contracts::ArrayOf#valid? refers to vals more than self (FeatureEnvy)
    Contracts::Bool has no descriptive comment (IrresponsibleModule)
    Contracts::Maybe#initialize refers to vals more than self (FeatureEnvy)
    Contracts::None#self.valid? has unused parameter 'val' (UnusedParameters)
    lib/contracts.rb -- 24 warnings:
    Contract#call_with calls _args[i] twice (DuplicateMethodCall)
    Contract#call_with has approx 13 statements (TooManyStatements)
    Contract#call_with has the variable name 'i' (UncommunicativeVariableName)
    Contract#call_with has the variable name 'j' (UncommunicativeVariableName)
    Contract#initialize calls contracts[-1] 3 times (DuplicateMethodCall)
    Contract#pretty_contract doesn't depend on instance state (UtilityFunction)
    Contract#pretty_contract has the parameter name 'c' (UncommunicativeParameterName)
    Contract#pretty_contract refers to c more than self (FeatureEnvy)
    Contract#self.failure_msg calls data[:contract] 4 times (DuplicateMethodCall)
    Contract#self.failure_msg calls data[:contract].to_s twice (DuplicateMethodCall)
    Contract#self.failure_msg calls data[:method] 7 times (DuplicateMethodCall)
    Contract#self.failure_msg has approx 9 statements (TooManyStatements)
    Contract#self.make_validator calls lambda 7 times (DuplicateMethodCall)
    Contract#self.make_validator contains iterators nested 2 deep (NestedIterators)
[...]

PR #117, opened 2015-03-29
Adit/rubocop

egonSchiele:
Fixes all rubocop warnings (issue #21).

I disabled some cops, you can see which ones and my rationales in `rubocop.yml`. Some big changes in this PR:

1. change function name `Invariant` to `invariant` (cc @waterlink). I can see why this was named to be consistent with the `Contract` decorator. But since this is not a decorator, I think it makes more sense for it to be lowercase (consistent with methods like `attr_accessor`).

2. Adding rubocop to travis-ci. It would be nice to follow a coding standard but I don't always catch every minor issue. TBH I also don't like discussing minor issues like that...it feels like bike-shedding. This way, there is one coding standard and it is enforced automatically.

Another reason for doing this: cleaning up code like I did with this PR, changes the history. If a line looks weird and someone does a git blame, they will be taken to this rubocop commit. Not the original commit that might have explained the code. So ideally I don't want to make separate commits like this for code cleanup...and that won't happen if we have rubocop on ci.

What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,4 @@ rvm:
 script:
   - bundle exec rspec
   - ruby rubocop_test.rb
-bundler_args: --without development
+bundler_args: --without development ruby2
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,6 @@ rvm:
   - jruby-19mode # JRuby in 1.9 mode
 
 script:
-  bundle exec rspec
-  bundle exec rubocop --config rubocop.yml
+  - bundle exec rspec
+  - bundle exec rubocop --config rubocop.yml
 bundler_args: --without development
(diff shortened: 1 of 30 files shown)


alex-fedorov:
:+1:

waterlink:
Build is failing because you need to add `rubocop` to `Gemfile`

waterlink:
Actually I tried to run locally, and I have lots and lots of errors, and even one rubocop bug error.. https://gist.github.com/waterlink/c91e168dc907e5f96490 Is it expected? Or do you have it green locally?

egonSchiele:
Did you run with the config file?

waterlink:
Yes, the command: `bundle exec rubocop --config rubocop.yml > /tmp/rubocop.log 2> /tmp/rubocop.err`. If it matters, I'm on ruby 2.1.3, which ruby version you are using?

waterlink:
Could be we have different versions of rubocop too..

```
bundle exec rubocop --version
0.23.0
```

alex-fedorov:
This may be of some help: https://github.com/rspec/rspec-core/blob/master/Gemfile#L32

egonSchiele:
Testing that out, thanks

PR #118, opened 2015-03-29
Fix useful_to_s? on older Rubies < 2.0

sfcgeorge:
`useful_to_s?` wasn't useful on Ruby < 2.0 as `#inspect` simply called `#to_s` so they were always the same. I've fixed that so it works properly in all Rubies.

### DIFF
--- a/lib/contracts/formatters.rb
+++ b/lib/contracts/formatters.rb
@@ -16,7 +16,7 @@ module Contracts
         elsif contract.is_a?(Array)
           array_contract(contract)
         else
-          InspectWrapper.new(contract, @full)
+          InspectWrapper.create(contract, @full)
         end
       end
 
@@ -24,32 +24,23 @@ module Contracts
       def hash_contract(hash)
         @full = true # Complex values output completely, overriding @full
         hash.inject({}) do |repr, (k, v)|
-          repr.merge(k => InspectWrapper.new(contract(v), @full))
+          repr.merge(k => InspectWrapper.create(contract(v), @full))
         end.inspect
       end
 
       # Formats Array contracts.
       def array_contract(array)
         @full = true
-        array.map { |v| InspectWrapper.new(contract(v), @full) }.inspect
+        array.map { |v| InspectWrapper.create(contract(v), @full) }.inspect
       end
     end
 
     # A wrapper class to produce correct inspect behaviour for different
     # contract values - constants, Class contracts, instance contracts etc.
-    class InspectWrapper
-      # Turn InspectWrapper into a factory, will never be an instance
-      class << self
-        alias_method :__new__, :new
-
-        def inherited(subclass)
-          class << subclass
-            alias_method :new, :__new__
-          end
-        end
-      end
-
-      def self.new(value, full = true)
+    module InspectWrapper
+      # InspectWrapper is a factory, will never be an instance
+      # @return [ClassInspectWrapper, ObjectInspectWrapper]
+      def self.create(value, full = true)
         if value.class == Class
           ClassInspectWrapper
         else
@@ -116,7 +107,9 @@ module Contracts
       end
     end
 
-    class ClassInspectWrapper < InspectWrapper
+    class ClassInspectWrapper
+      include InspectWrapper
+
       def custom_to_s?
         @value.to_s != @value.name
       end
@@ -126,7 +119,9 @@ module Contracts
       end
     end
 
-    class ObjectInspectWrapper < InspectWrapper
+    class ObjectInspectWrapper
+      include InspectWrapper
+
       def custom_to_s?
         !@value.to_s.match(/#\<\w+:.+\>/)
       end
--- a/lib/contracts/formatters.rb
+++ b/lib/contracts/formatters.rb
@@ -85,12 +85,19 @@ module Contracts
 
       def useful_to_s?
         # Useless to_s value or no custom to_s behavious defined
-        @value.to_s != "" &&
-          if @value.class == Class # It's a class contract
-            @value.to_s != @value.name
-          else # It's an instance contract
-            !@value.to_s.match(/#\<\w+:.+\>/)
-          end
+        !empty_to_s? && custom_to_s?
+      end
+
+      def empty_to_s?
+        @value.to_s.empty?
+      end
+
+      def custom_to_s?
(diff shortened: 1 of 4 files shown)


lib/contracts/formatters.rb
```diff
@@ -85,8 +85,12 @@ def plain?
 
       def useful_to_s?
         # Useless to_s value or no custom to_s behavious defined
-        # Ruby < 2.0 makes inspect call to_s so this won't work
-        @value.to_s != '' && @value.to_s != @value.inspect
+        @value.to_s != '' &&
+          if @value.class == Class # It's a class contract
```
waterlink: multiline `if..else..end` as a second condition to complex condition after `&&` looks confusing to me. Any chance extracting `if..else..end` part to method?

sfcgeorge: Hah you're right, I'll see what I can do.

sfcgeorge:
Phew, that was a pain to fix. I'm still not convinced it's entirely correct but at least all current tests and the one I added are passing. 

inspect calls to_s in old Rubies so we can't compare them to see if to_s is useful. Instead, if it's a class I check to see if to_s is the same as the class name, if it's an object I check to see if to_s looks like an object representation `#<Class:...>`.

But then if it wasn't useful... we were calling inspect, which in those old Rubies did the same thing as to_s, so we were substituting a useless to_s call with a useless inspect call. I wrote a raw_inspect that tries to do something better. If to_s isn't empty then inspect is called and the result will vary between Rubies. If to_s is empty (specific case I was trying to fix) then it calls name instead. This will look right for classes but won't look right for objects because it isn't object representation. I'm open to changes of behaviour in this method. The basic requirement is that it doesn't print an empty string if to_s is empty string.

waterlink:
LGTM

lib/contracts/formatters.rb
```diff
@@ -85,8 +85,28 @@ def plain?
...
-        @value.to_s != "" && @value.to_s != @value.inspect
+        !empty_to_s? && custom_to_s?
+      end
+
+      def empty_to_s?
+        @value.to_s.empty?
+      end
+
+      def custom_to_s?
+        if @value.class == Class # It's a class contract
```
waterlink:
```ruby
if @value.class == Class
  # ...
else
  # ...
end
```

This construction duplicated. Not that it is so bad, but could be better:

```ruby
class InspectWrapper
  def lifted_class
    @value.class == Class ? ClassContractInspectWrapper : InstanceContractInspectWrapper
  end

  def lift
    lifted_class.new(@value)
  end

  def custom_to_s?
    lift.custom_to_s?
  end

  def raw_inspect
    lift.raw_inspect.gsub(/^Contracts::/, "")
  end
end

class InstanceContractInspectWrapper
  def custom_to_s?
    !@value.to_s.match(/#\<\w+:.+\>/)
  end

  def raw_inspect
    empty_to_s? ? @value.class.name : @value.inspect
  end
end

class ClassContractInspectWrapper
  def custom_to_s?
    @value.to_s != @value.name
  end

  def raw_inspect
    empty_to_s? ? @value.name : @value.inspect
  end
end
```

More complicated, but more DRY and robust.

waterlink: Don't really know what is better in that case. So for you @sfcgeorge to decide :)

sfcgeorge:
To reduce duplication and allow for speciffic class vs object inspect behaviour I did something similar to what @waterlink suggested and turned InspectWrapper into a factory. I wondered if `self.new` can be overridden, and it can, it just requires a little alias trick. I love this, I think it's equal parts genius and madness. If you think it's too far on the mad side I can do something more sane though.

alex-fedorov:
Looks good to me, WDYT @egonSchiele on this?

egonSchiele:
I don't like overriding `new`. I would just make a method on `InspectWrapper` called `build` or something. Less patching == less chance of running into obscure edge cases.

sfcgeorge:
I found a simpler less magic way to override `new`. I like this abstract factory approach best because it leaves the implementation of InspectWrapper hidden, and you don't have to keep track of which objects have to be instantiated with `new` and which with `create`.

If you don't like the use of `allocate` then I can change InspectWrapper to a module and then there's no problem defining `new` on it, then `ClassInspectWrapper` and `ObjectInspectWrapper` would include it rather than inheriting from it.

Or if you really don't like overriding `new` however it's done then I can change it to `create`.

egonSchiele:
I don't like the idea of overriding `new`. My issues are:

- Unexpected behavior: When I call `InspectWrapper.new`, I expect to get an instance of `InspectWrapper`.
- Unnecessary complexity. I don't know all the consequences that this has. It's fine for the use case we have right now. But maybe down the line we will add more functionality to the wrapper classes, and run into a subtle bug with this. I don't want to have to deal with that.

I think we should make it explicit that `InspectWrapper` is a factory, so I think using `create` instead of `new` is a good thing.

One thing I like about this implementation: it means `InspectWrapper` is totally abstract, you can't accidentally make an instance of `InspectWrapper`. But I don't think it's worth trading off the issues I listed.

sfcgeorge:
Ok I'll change to `create`, and fix the Rubocop mistakes while I'm at it.

sfcgeorge:
I've removed the overriding `new` in favour of `create`. I've seen the create naming used in other gems.

I have made InspectWrapper a module and given it a class method `new` that raises an exception explaining that it's a factory and `create` must be used. It helped me in changing all the old calls to `new`, if you don't think it's necessary I can quickly remove it.

nixpulvis:
I'm a bit disconnected from this discussion, however seems like a module would have fit the bill better than this "abstract class" since Ruby doesn't really have abstract classes.

sfcgeorge:
Not strict ones no, it's conceptually abstract :) But yeah I suppose that is a reason to use a module, good thought. 

That said, a fake `new` method is now unnecessary as modules can't be instantiated so I've removed it. I did the change in 2 steps and hadn't fully thought it through. Simpler now.

nixpulvis:
I'm not necessarily saying you should rewrite it (not my place), but just a thought for your future Ruby adventures.

sfcgeorge:
Ugh I broke it, sorry for all the revisions. I've fixed it and rebase squashed the last few commits together.

egonSchiele:
thanks!

PR #120, opened 2015-04-01
Resolve issue with singleton classes

actsasbuffoon:
Including Contracts into a class can result in a StackLevelTooDeep under certain (somewhat mysterious) circumstances. The issue appears to be related to the fact that Ruby provides a `singleton_class?` method, and it appears that ActiveSupport does as well. The implementations differ in subtle ways.

ActiveSupport's implementation detects singleton classes by checking to see if they aren't the first entry in their array of ancestors. Singleton classes are sometimes called ghost classes since they don't
ordinarily show up in calls to `ancestors`, despite the fact that they sit in the method lookup chain.

Despite all of that, it appears that sometimes singleton classes *do* show up in calls to `ancestors`. My suspicion is that some reasonably common gem is being overly aggressive with monkey patching, but I
haven't figured out who the culprit is yet.

After discussing it with Contracts maintainers, @waterlink suggested adding some helper methods to the Support module that should make Contracts more resilient in the face of a misbehaving implementation of `ancestors`.


waterlink:
Heh, rubocop doesn't like the name `is_eigenclass?` and tells to remove `is_` part. Fine with me.

egonSchiele:
nice to see rubocop is working :+1:

PR #122, opened 2015-04-02
Rename rubocop.yml to .rubocop.yml for automatic running

sfcgeorge:
Rubocop will automatically load a config file if it is called `.rubocop.yml` so I renamed it to that. 

Advantages:

* Rubocop run automatically by Syntastic in Vim will use the config file now like magic. Before it didn't so I'd see incorrect errors.
* Now it matches `.travis.yml` style.


nixpulvis:
:+1:

egonSchiele:
Didn't know about this, thanks!

Issue #104, opened 2015-03-11
Inherited contracts


nixpulvis:
This is a pretty common use case (for me at least). I define a base class:

```ruby
class Base
  # ...
  Contract String => String
  def foo(arg)
    # ...
  end
  # ...
end
```

Next I write a bunch of subclasses from `Base`. I might write a bunch of new methods on these subclasses, but sometimes I also need to overwrite the definition of a function.

```ruby
class A < Base
  # ...
  def foo(arg
    # New stuff.
  end
  # ...
end
```

By default it would be ideal for the contract of `Base#foo` to be given to `A#foo`, allowing a new contract to overwrite the inherited one if needed.

egonSchiele:
Hmm, I don't agree with this. One reason I like contracts, the contract is right above the method definition. So I don't have to go hunting to understand what a method is doing. If you see a method with no contract on it, and then it throws a contract error, that could get confusing.

I can see this being an issue when you are overriding a bunch of one line methods. But I would rather have that be more verbose for the sake of clarity.

nixpulvis:
Maintaining interfaces is the **most** important thing this library provides. If I need to duplicate an contract then I need to maintain `n` contracts for the same interface, where `n` is the number of implementors of the interface.

The blaming feature should be able to handle pointing contract failure to the correct location.

stevesun21:
I have the question, why the contracts I wrote on the "interface" cannot be inherited? 

So far, I have to use proxy pattern to solve this problem.
```ruby
class Base
  # ...
  Contract String => String
  def foo(arg)
    # ...
  end

  def proxy(client)
    @client = client
    self
  end
  # ...
end

usage:

obj = Base.new.proxy(A.new).foo(...)
```

waterlink:
There is one solution: stop using inheritance for code reuse - it only increases coupling and calls more problems. Use composition instead.

waterlink:
There was no so much activity here, and I am reluctant to closing this because @egonSchiele thinks that it is important to spell out contracts explicitly for each method.

WDYT?

nixpulvis:
While I agree with you in general there are two reasons I see this as important nonetheless.

1. People do not all agree that they should not use inheritance, and given that Ruby explicitly supports it people will use it.
2. Enforcing contracts on subclasses actually makes inheritance a better option in some cases. It adds structure to interfaces developed for reuse.

sebnozzi:
[...]

PR #125, opened 2015-04-03
Receiver contract

actsasbuffoon:
This adds a receiver contract which automatically narrows when inherited.

### Example ###

Let's say we've got a class hierarchy that goes something like this:

```ruby
class BaseCounter
  Contract None => Num
  attr_accessor :value

  Contract Num => Any
  def initialize(initial_value)
    @value = initial_value
  end

  Contract BaseCounter => BaseCounter
  def +(other)
    self.class.new(value + other.value)
  end
end

class RedCounter < BaseCounter; end
class BlueCounter < BaseCounter; end
```

With these classes, we could do something like this:

```ruby
red1 = RedCounter.new 3
red2 = RedCounter.new 5

red1 + red2
# => #<RedCounter @value=8>
```

That's pretty nifty, but we run into a problem:

```ruby
red = RedCounter.new 3
blue = BlueCounter.new 5

red + blue
# => #<RedCounter @value=8>
```

When we created the class hierarchy, we didn't want red and blue counters to mix. They've got the same parent class, but that doesn't mean that we wanted them to be interchangeable.

This PR adds a new kind of contract that handles this use case.

```ruby
class BaseCounter
  Contract None => Num
  attr_accessor :value

  Contract Num => Any
  def initialize(initial_value)
    @value = initial_value
  end

  # Note that these are Receiver contracts now, instead of BaseCounter
  Contract Receiver => Receiver
  def +(other)
    self.class.new(value + other.value)
  end
end

class RedCounter < BaseCounter; end
class BlueCounter < BaseCounter; end

red = RedCounter.new 3
blue = BlueCounter.new 5

red + blue
# => Contract violation for argument 1 of 1: (ContractError)
# ~>         Expected: Receiver,
# ~>         Actual: #<BlueCounter @value=5>
# ~>         Value guarded in: BaseCounter::+
# ~>         With Contract: Receiver => Receiver
```

### Notes ###

This contract didn't lend itself particularly well to the way Contracts is structured. I had to add a receiver stack (implemented to be thread and fiber safe), and perform arity checking to see if the receiver data should be passed along.

My biggest complaint about this contract is that it doesn't print out very nicely. Rather than `Expected: Receiver`, it would be nice to see `Expected: RedCounter`. Once again, it would require a fair degree of refactoring the contract inspection code to make that work well, so I'm leaving it alone at the moment.

I'd be happy to add some documentation for this contract, but I wanted to see how people feel about this contract first.

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -70,7 +70,10 @@ class Contract < Contracts::Decorator
     fail data[:contracts].failure_exception.new(failure_msg(data), data)
   end
 
+  Thread.current[:contracts_receivers] = []
+
   attr_reader :args_contracts, :ret_contract, :klass, :method
+
   def initialize(klass, method, *contracts)
     if contracts[-1].is_a? Hash
       # internally we just convert that return value syntax back to an array
@@ -239,7 +242,9 @@ class Contract < Contracts::Decorator
     else
       # classes and everything else
       # e.g. Fixnum, Num
-      if contract.respond_to? :valid?
+      if contract.respond_to?(:valid?) && contract.method(:valid?).arity == 2
+        lambda { |arg| contract.valid? arg, :receiver => Thread.current[:contracts_receivers].last }
+      elsif contract.respond_to? :valid?
         lambda { |arg| contract.valid?(arg) }
       elsif klass == Class
         lambda { |arg| arg.is_a?(contract) }
@@ -280,6 +285,7 @@ class Contract < Contracts::Decorator
   end
 
   def call_with(this, *args, &blk)
+    Thread.current[:contracts_receivers] << this
     args << blk if blk
 
     # Explicitly append blk=nil if nil != Proc contract violation anticipated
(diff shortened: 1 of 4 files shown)


nixpulvis:
Contracts and inheritance is a hard problem which is going to require some overarching design if we want to tackle it see #104. I don't fully understand the general use case for this builtin contract, or at least the naming and examples seem misleading.

Generally speaking the receiver is the object which a message is passed to, for example in `foo.bar` foo is the receiver.

I do strongly sugest we start thinking about inheritance, and contracts.

egonSchiele:
This is a neat contract to have, but I'm nervous that we haven't thought of all the possible corner cases with it. Are you using it in your code? If so it would be great to use that as a test bed, to see how this contract works, before adding it to the library.

waterlink:
I don't really like the implementation (precisely: `Thread.current` usage).

But it is nice contract to have, especially for these cases when you need to make sure that the argument is of the same type as a self.

What we can actually do, is to always pass `self` to a contract that responds to `:valid_with_receiver?`. This functionality can be added in `lib/contracts/validators.rb`

waterlink:
@actsasbuffoon Are you still up to updating this PR? Or it can be closed?

waterlink:
@actsasbuffoon I am closing it because of lack of any activity. If needed re-open please.

PR #128, opened 2015-04-10
Change syntax for no arguments

justincampbell:
In many languages, a type signature for a function that increments a number by one looks like `Num -> Num`, which is represented fine in this library as `Contract Num => Num`. But for a function that just returns a value, say pi: What is pi? Pi is a Num (or Float/BigDecimal/whatever). Pi is not a `nil => Num`. I think it would improve the readability of the contract type signatures if there was a way to express the type of a declared value or computed value as a single type, and omit the `nil =>`.

This change allows you to write:

```rb
Contract Num
def pi
  3.14
end
```

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -78,16 +78,21 @@ class Contract < Contracts::Decorator
 
   attr_reader :args_contracts, :ret_contract, :klass, :method
   def initialize(klass, method, *contracts)
-    if contracts[-1].is_a? Hash
-      # internally we just convert that return value syntax back to an array
-      @args_contracts = contracts[0, contracts.size - 1] + contracts[-1].keys
-      @ret_contract = contracts[-1].values[0]
-    else
-      fail %{
-        It looks like your contract for #{method} doesn't have a return value.
-        A contract should be written as `Contract arg1, arg2 => return_value`.
-      }.strip
+    unless contracts.last.is_a?(Hash)
+      unless contracts.one?
+        fail %{
+          It looks like your contract for #{method} doesn't have a return
+          value. A contract should be written as `Contract arg1, arg2 =>
+          return_value`.
+        }.strip
+      end
+
+      contracts = [nil => contracts[-1]]
     end
+
+    # internally we just convert that return value syntax back to an array
+    @args_contracts = contracts[0, contracts.size - 1] + contracts[-1].keys
+    @ret_contract = contracts[-1].values[0]
 
     @args_validators = args_contracts.map do |contract|
       Contract.make_validator(contract)
--- a/lib/contracts/decorators.rb
+++ b/lib/contracts/decorators.rb
@@ -93,14 +93,14 @@ module Contracts
       # Here's why: Suppose you have this code:
       #
       #     class Foo
-      #       Contract nil => String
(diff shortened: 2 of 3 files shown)


nixpulvis:
What about `None => ...`.

justincampbell:
@nixpulvis IMO that's the same as `nil => ...`, just a different placeholder

NobbZ:
How can one differ between having one argument and returning nil, just for
the sake of side-effects and having no arguments but a return type after
your proposal? I like the way I have to mention either way explicitely as
it was in C the earlier days `void foo (Int a)` or `Int bar (void)`.
Am 10.04.2015 14:04 schrieb "Justin Campbell" <notifications@github.com>:

egonSchiele:
Nobbz, could you expand on what you mean? You can write a contract for one argument and return nil like this:

    Contract String => nil

You can write "no arguments but a return type like this:"

    Contract nil => String # current way
    Contract String # proposed way

waterlink:
@egonSchiele I think @NobbZ means this case

```ruby
# I don't care here about return value.
Contract String
def puts(value)
  io.writeln(value)
end
```

waterlink:
@justincampbell Actually, `Contract None => ..` is not the same as `Contract nil => ..`. Consider these examples:

```ruby
Contract None => ...
def do_something
end

do_something(nil)    #=> ContractError
```

```ruby
Contract nil => ...
def do_something
end

do_something(nil)    #=> ArgumentError
```

egonSchiele:
@waterlink @NobbZ ok, I can see how it would be confusing. The problem is:

1. It's unclear how functions with no arguments, and functions with no return value, are supposed to be handled.
2. The current solution for functions with no arguments is pretty unelegant.

Haskell solves this elegantly. Everything is a function in Haskell. So for example, these are all functions:

```haskell
add :: Int -> Int -> Int
add a b = a + b

double :: Int -> Int
double x = x * 2 

five :: Int
five = 5
```

The first function takes two `Int`s and returns an `Int`.

The second function takes an `Int` and returns an `Int`.

The third function takes no args and returns an `Int`.

You can see how the syntax that @justincampbell is proposing, follows naturally in Haskell. So I propose adopting this convention for contracts. For functions with no params:

```ruby
Contract Int
def five
  5
end
```

And if you don't care about the return value:

```ruby
Contract String => Any
def do_something str
  ...
end
```

- Adopting a convention would solve problem #1. Plus if we have users with a Haskell background, this will be the syntax they are used to.
- `Contract Int` looks a lot cleaner than `Contract nil => Int`, solving #2.

One issue as @nobbz pointed out is this version is less explicit. But I think that's fine...it has never been a problem for me in Haskell. Another issue is, not everyone is familiar with Haskell :)

egonSchiele:
I would add these tests though @justincampbell 

- Check for a bad contract like `Contract String, String`. That should raise an error.
- You shouldn't be able to write `Contract String` if the method takes one or more arguments.

egonSchiele:
Merged since no one had any objections. Thanks for the PR.

justincampbell:
Thanks! Sorry I didn't get around to adding those tests. It looks like they were added on master. :tada:

PR #129, opened 2015-04-10
Ignore Gemfile.lock

justincampbell:
Having the lockfile actually prevented me from being able to `bundle install`. I had to do a `bundle update` instead (which changed the `Gemfile.lock`). Commiting a `Gemfile.lock` also increases the risk of library incompatibilities, since people installing the gem from RubyGems with `gem install contracts` or putting `gem 'contracts'` in their `Gemfile` will not respect the `Gemfile.lock`.

For more information on when and when not to commit a `Gemfile.lock`, see http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/


nixpulvis:
:+1:

waterlink:
Hmm, this introduced rubocop failure. Probably because rubocop got updated..

egonSchiele:
Thank you, I didn't know about this difference. Rubocop did get updated. If you change the relevant line in the Gemfile to:

    gem "rubocop", "~> 0.29.1", :platform => [:ruby_20, :ruby_21]

The build should be fixed.

justincampbell:
@egonSchiele done!

egonSchiele:
Thank you!

PR #134, opened 2015-04-12
Add SetOf builtin contract

waterlink:
To me this contract looks generic enough to be included as builtin contract. What do you think?

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -259,6 +259,29 @@ module Contracts
     end
   end
 
+  # Takes a contract. The related argument must be a set.
+  # Checks the contract against every element of the set.
+  # If it passes for all elements, the contract passes.
+  # Example: <tt>SetOf[Num]</tt>
+  class SetOf < ArrayOf
+    def valid?(vals)
+      lazy_require_set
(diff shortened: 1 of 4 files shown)


lib/contracts/builtin_contracts.rb
```diff
@@ -259,6 +259,29 @@ def to_s
...
 
+  # Takes a contract. The related argument must be a set.
+  # Checks the contract against every element of the set.
+  # If it passes for all elements, the contract passes.
+  # Example: <tt>SetOf[Num]</tt>
+  class SetOf < ArrayOf
+    def valid?(vals)
+      lazy_require_set
+      return false unless vals.is_a?(Set)
+      super(vals.to_a)
```
egonSchiele: Hidden memory allocation here with `to_a`...I would just use `vals.each`.

egonSchiele:
Very nice idea, just don't like the hidden performance/memory penalty^

alex-fedorov:
@egonSchiele removed memory penalty by extracting common abstraction to private `CollectionOf` contract and made `ArrayOf` and `SetOf` just specific factories of `CollectionOf` contract

waterlink:
Rebased against green master, and this one should become green too

PR #135, opened 2015-04-12
Make HashOf contract prettier

waterlink:
Adds ability to use `HashOf` in this manner:

```ruby
Contract HashOf[KeyContract => ValueContract] => ...
```

WDYT?

btw, old syntax is still preserved, ie: `HashOf[KeyContract, ValueContract]`


egonSchiele:
Great idea! I would use this as the example in TUTORIAL.md and remove the old style from the tutorial entirely. It's nice that you are still supporting the old style though.

Can you also have it throw an error if the hash has more the one key? Or maybe just print a warning.

alex-fedorov:
@egonSchiele, done as suggested:

- Remove old syntax from tutorial.
- Raise argument error when `HashOf` gets multiple key-value pairs in definintion

waterlink:
Rebased against green master, and this one should become green too

waterlink:
Rebased again..

egonSchiele:
Thanks!

PR #136, opened 2015-04-13
Fix failing fixture for old rubies

waterlink:


alex-fedorov:
Ah, if here would not be enough, so that means putting them into ruby_specific folders

waterlink:
This should make the build green again. /cc @egonSchiele

egonSchiele:
Thank you!

PR #141, opened 2015-04-28
Refactor to contracts engine

### LINKED ISSUE #138: Decorator implementation conflicts with Draper
Both implementations define `decorate`. 

It seems that Contract's decorators are intended only for internal consumption and could be hidden away. One strategy is to have a single "engine" variable which is composed onto a class. The engine contains all data and methods required by Contracts and effectively provides a namespace, preventing any instance variables and methods from conflicting with those defined by the class. All functionality is accessed through the engine instance which contains a reference to the owning class such that `object.decorate` becomes `object.contracts.decorate`.

waterlink:
Allows to be as non-intrusive as possible, moving all the functionality effectively to only one method `.__contracts_engine` (which has rich object(s) inside).

Still needs some cleanup:

- [x] refactor `MethodHandler#handle`.
- [x] remove `Eigenclass`.
- [x] remove `include Contracts::Module` since now they get all the needed stuff properly.
- [x] try to reduce amount of public methods on `Engine`. Or at least mark some of them as `_pseudo_private`. There is a probability, that some of them could be moved to protected (interaction between `Engine` and `EigenclassEngine`).
- [x] add documentation in new code

closes #138


waterlink:
@egonSchiele @sfcgeorge This one is not finished yet, but can you take a look and tell if you like the direction where this PR is going?

lib/contracts/engine.rb
```diff
@@ -0,0 +1,142 @@
...
+    def decorate(klass, *args)
+      validate!
+      decorators << [klass, args]
+    end
+
+    def validate!
+    end
+
+    def set_eigenclass_owner
+      Engine.fetch_from(eigenclass).owner_class = target
```
waterlink: just `eigenclass_engine` here

lib/contracts/modules.rb
```diff
@@ -10,8 +10,8 @@ def self.extended(base)
 
     def self.common(base)
       return unless base.instance_of?(Module)
-      base.extend(MethodDecorators)
-      Eigenclass.lift(base)
+      #base.extend(MethodDecorators)
```
waterlink: So this one became a no-op. Effectively means, that `Contracts::Modules` can be safely removed.

betamatt:
:heart:

egonSchiele:
Works for me. Also happy that Contracts::Module can be removed! Since you are working on this, could you please add comments on all the methods? That will make it easy to modify this code next time...it is pretty complex.

waterlink:
Finally green on CI. @egonSchiele can you take a look?

egonSchiele:
Looks great. Thank you for adding all the comments!

egonSchiele:
Weirdly, the change did make contracts a little slower.

Old:

    testing add                      0.430000   0.000000   0.430000 (  0.435320)
    testing contracts add            3.540000   0.000000   3.540000 (  3.547737)

New:

    testing add                      0.430000   0.000000   0.430000 (  0.435508)
    testing contracts add            3.950000   0.000000   3.950000 (  3.960544)

Not a meaningful slowdown.

PR #142, opened 2015-04-28
Fix maybe proc

### LINKED ISSUE #139: Combining Args and Optional Proc fails
Example, which is the signature Rack uses for middleware:

```
Contract Any, Args[Any], Maybe[Proc] => Any
def use(middleware, *args, &block)
  @middleware << MiddlewareDescriptor.new(middleware, args, block)
end
```

In 0.9, the Args assumes the last value is either Proc or nil. But when no proc is passed, the last arg should be a part of the `*args`. Perhaps `Maybe` is not right for something so weirdly Ruby as the optional block usage.

In code:

```
use ::Rack::Static, path: "/"
```

or

```
use ::MinifyCSS do |m|
  m.compressor = :yahoo
end
```

sfcgeorge:
Fixes #139

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -113,8 +113,9 @@ class Contract < Contracts::Decorator
     # == @has_proc_contract
     last_contract = args_contracts.last
     is_a_proc = last_contract.is_a?(Class) && (last_contract <= Proc || last_contract <= Method)
+    maybe_a_proc = last_contract.is_a?(Contracts::Maybe) && last_contract.include_proc?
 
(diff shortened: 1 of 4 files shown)


sfcgeorge:
Before we weren't checking for `Maybe[Proc]` in the `has_proc_contract` test. Simple fix, though the code is a bit edge-casey. I can't think of a better way without completely rewriting the arg parsing code again.

waterlink:
looks good enough to me

egonSchiele:
Yay, thank you! Merged.

PR #147, opened 2015-05-13
Fix Pos,Neg contracts which fail incorrectly for nil values

betamatt:
Pos and Neg contracts explode when passed nil values. This is most easily noticed when combined with Maybe -  `Maybe[Pos]` - which will error when a positive value is not provided.

For example:
```
  1) Contracts: Pos: should fail for nil
     Failure/Error: expect { @o.pos_test(nil) }.to raise_error(ContractError)
       expected ContractError, got #<NoMethodError: undefined method `>' for nil:NilClass>
```


egonSchiele:
Thanks!

PR #151, opened 2015-05-19
Add KeywordArgs and Optional contracts

### LINKED ISSUE #131: Problem with default keyword arguments
The following code generates the exception shown underneath.

    require 'contracts'
    include Contracts

    Contract ({nfiles:  Nat}) => ArrayOf[String]
    def list(nfiles: 10)
      []
    end

    list(nfiles: 4)      # No problem here.
    list()               # Causes exception

Exception:

    /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts/builtin_contracts.rb:70:in `valid?': undefined method `>=' for nil:NilClass (NoMethodError)
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:249:in `block in make_validator'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:207:in `[]'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:207:in `valid?'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:234:in `block (2 levels) in make_validator'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:233:in `each'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:233:in `all?'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:233:in `block in make_validator'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:303:in `[]'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:303:in `block in call_with'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:298:in `times'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts.rb:298:in `call_with'
      from /Library/Ruby/Gems/2.0.0/gems/contracts-0.8/lib/contracts/decorators.rb:149:in `block in common_method_added'
      from /Users/Gavin/tmp/contracts-bug.rb:23:in `<main>'

I'm calling it a bug because it doesn't give a contract violation message, rather an internal exception in the contracts library itself.

I've not used keyword arguments much, though. Am I right to expect that this code should work?

waterlink:
Fixes #131 

Allows to do this:

```ruby
Contract OptHash[nfiles: Opt[Nat], mfiles: Nat] => ArrayOf[String]
def list(nfiles: 10, mfiles:)
  []
end

# both of these work as expected
list(nfiles: 4, mfiles: 77)
list(mfiles: 77)

# and this will fail with contract error as expected as opposed to `Maybe[Nat]` solution
list(mfiles: 77, nfiles: nil)
```

Contract names are subject to discussion.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -81,6 +81,8 @@ contracts.ruby comes with a lot of built-in contracts, including the following:
 * [`ArrayOf`](http://www.rubydoc.info/gems/contracts/Contracts/ArrayOf) – checks that the argument is an array, and all elements pass the given contract, e.g. `ArrayOf[Num]`
 * [`SetOf`](http://www.rubydoc.info/gems/contracts/Contracts/SetOf) – checks that the argument is a set, and all elements pass the given contract, e.g. `SetOf[Num]`
 * [`HashOf`](http://www.rubydoc.info/gems/contracts/Contracts/HashOf) – checks that the argument is a hash, and all keys and values pass the given contract, e.g. `HashOf[Symbol => String]`
+* [`OptHash`](http://www.rubydoc.info/gems/contracts/Contracts/OptHash) – checks that the argument is an options hash, and all required keyword arguments are present, and all values pass their respective contracts, e.g. `OptHash[:number => Num, :description => Opt[String]]`
+* [`Opt`](http://www.rubydoc.info/gems/contracts/Contracts/Opt) – checks that the keyword argument is either not present or pass the given contract, can not be used outside of `OptHash` contract, e.g. `Opt[Num]`
 * [`Maybe`](http://www.rubydoc.info/gems/contracts/Contracts/Maybe) – passes if the argument is `nil`, or if the given contract passes
 * [`RespondTo`](http://www.rubydoc.info/gems/contracts/Contracts/RespondTo) – checks that the argument responds to all of the given methods, e.g. `RespondTo[:password, :credit_card]`
 * [`Send`](http://www.rubydoc.info/gems/contracts/Contracts/Send) – checks that all named methods return true, e.g. `Send[:valid?]`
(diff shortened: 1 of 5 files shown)


nixpulvis:
I'd call the contract `Options` personally.

waterlink:
@nixpulvis thought about that too, but the chance that the user will have the `Options` class in their domain is much higher (just my gut feeling, maybe not)

nixpulvis:
Yea, I actually hit this with existing contracts pretty often anyway. I think the real solution is namespacing these contracts under something like `C::Options`.

waterlink:
So how about KeywordArgs name then? WDYT?

TUTORIAL.md
```diff
@@ -239,6 +241,52 @@ give_largest_value(a: 1, b: 2, c: 3) # returns 3
...
+
+But this will not quite work if you want to have a default values:
+
+```ruby
+Contract String, { :port => Num, :user => String, :password => String } => Connection
+def connect(host, port: 5000, user:, password:)
+```
+
+```
+ContractError: Contract violation for argument 2 of 2:
```
egonSchiele: Just before this, can you add a line that calls the `connect` function that will throw the related ContractError, and make it obvious that we are not passing in a value for `port`?

lib/contracts/builtin_contracts.rb
```diff
@@ -345,6 +345,77 @@ def validate_hash(hash)
...
+    end
+
+    def valid?(hash)
+      options.all? do |key, contract|
+        Opt._valid?(hash, key, contract)
+      end
+    end
+
+    def to_s
+      "OptHash[#{options}]".gsub(/Contracts::/, "")
```
egonSchiele: You shouldn't need to strip the `Contracts::` prefix, the formatter should do that for you automatically: https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts/formatters.rb#L105

Is that not working?

spec/ruby_version_specific/contracts_spec_2.1.rb
```diff
@@ -1,6 +1,6 @@
 class GenericExample
-  Contract String, Bool, Args[Symbol], Float, { e: Range, f: Maybe[Num] }, Proc =>
-           [Proc, Hash, Num, Range, Float, ArrayOf[Symbol], Bool, String]
+  Contract String, Bool, Args[Symbol], Float, OptHash[e: Range, f: Opt[Num]], Proc =>
```
egonSchiele: Can you also add a test that checks to make sure it succeeds when no value for `f` is passed in?

egonSchiele:
Left comments, looks good otherwise. :+1: for updating the tutorial. `KeywordArgs` sounds better to me since that reflects where this contract will be used..

lib/contracts/builtin_contracts.rb
alex-fedorov: Somehow it doesn't, I would like to look why

alex-fedorov: Because it uses `^Contracts::` regex, which makes it expect `Contracts::` to be at the start of the line (which is not the case).

alex-fedorov: Used `Formatters::InspectWrapper` in `Opt` contract

spec/ruby_version_specific/contracts_spec_2.1.rb
alex-fedorov: It is already tested by this suite - look at the full file, it is the first example.

alex-fedorov:
@egonSchiele Addressed.

Plus renamed both contracts to `KeywordArgs[ ..a-hash.. ]` and `Optional[ ..a-contract.. ]`.

egonSchiele:
Thanks!

sfcgeorge:
Nice!

gsinclair:
Excellent!

If we're discussing names, my choice is NamedArgs and Opt.  I think brevity is a good aim: contract specs can get quite long.

egonSchiele:
Hey @gsinclair, we already merged this as `KeywordArgs`.

PR #153, opened 2015-05-24
Recognize module as a class contract

### LINKED ISSUE #152: Modules as contracts are recognised as value contract instead of class contract
Given:

```ruby
module M; end

class X
  include M
end

Contract M => nil
def hello(x)
end
```

Expected:

```ruby
hello(X.new)     # => nil
```

Actual:

```
ParamContractError: Contract violation for argument 1 of 1:
        Expected: M,
        Actual: #<X:0x00000102a2df98>
        Value guarded in: Object::hello
        With Contract: Module => NilClass
        At: (irb):10
```

Relevant code: https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts.rb#L252-L253

Condition should be `klass == Class || klass == Module`.

Why? - module inclusion is valid way to do inheritance, so I want to check if something is enumerable or comparable (and much more stuff I do :) ) for example.

waterlink:
fixes #152 

/cc @egonSchiele

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -249,7 +249,7 @@ class Contract < Contracts::Decorator
       # e.g. Fixnum, Num
       if contract.respond_to? :valid?
         lambda { |arg| contract.valid?(arg) }
-      elsif klass == Class
+      elsif klass == Class || klass == Module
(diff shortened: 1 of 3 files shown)


egonSchiele:
Looks good :+1:

PR #154, opened 2015-05-26
Rename "Support" module to "ContractsSupport"

guizmaii:
Hi,

I'm beginning to use your great gem with one of my (rails) project but I have a namespace collision with your "Support" module.

Is it possible to rename it to "ContractsSupport" as proposed in this PR ?

KR,
Jules

### DIFF
--- a/benchmarks/bench.rb
+++ b/benchmarks/bench.rb
@@ -44,7 +44,7 @@ def profile
   profilers << MethodProfiler.observe(Object)
   profilers << MethodProfiler.observe(Contracts::MethodDecorators)
   profilers << MethodProfiler.observe(Contracts::Decorator)
-  profilers << MethodProfiler.observe(Contracts::Support)
+  profilers << MethodProfiler.observe(Contracts::ContractsSupport)
   profilers << MethodProfiler.observe(UnboundMethod)
   10_000.times do |_|
     contracts_add(rand(1000), rand(1000))
--- a/benchmarks/invariants.rb
+++ b/benchmarks/invariants.rb
@@ -60,7 +60,7 @@ def profile
   profilers = []
   profilers << MethodProfiler.observe(Contract)
   profilers << MethodProfiler.observe(Object)
-  profilers << MethodProfiler.observe(Contracts::Support)
+  profilers << MethodProfiler.observe(Contracts::ContractsSupport)
   profilers << MethodProfiler.observe(Contracts::Invariants)
   profilers << MethodProfiler.observe(Contracts::Invariants::InvariantExtension)
   profilers << MethodProfiler.observe(UnboundMethod)
--- a/benchmarks/io.rb
+++ b/benchmarks/io.rb
@@ -39,7 +39,7 @@ def profile
   profilers << MethodProfiler.observe(Object)
   profilers << MethodProfiler.observe(Contracts::MethodDecorators)
   profilers << MethodProfiler.observe(Contracts::Decorator)
-  profilers << MethodProfiler.observe(Contracts::Support)
+  profilers << MethodProfiler.observe(Contracts::ContractsSupport)
   profilers << MethodProfiler.observe(UnboundMethod)
   10.times do |_|
     contracts_download(@urls.sample)
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -4,7 +4,7 @@ require "contracts/errors"
 require "contracts/formatters"
 require "contracts/invariants"
 require "contracts/method_reference"
-require "contracts/support"
+require "contracts/contracts_support"
 require "contracts/engine"
 require "contracts/method_handler"
 
@@ -136,8 +136,8 @@ class Contract < Contracts::Decorator
   # and uses the hash passed into the failure_callback method.
   def self.failure_msg(data)
     expected = Contracts::Formatters::Expected.new(data[:contract]).contract
-    position = Contracts::Support.method_position(data[:method])
-    method_name = Contracts::Support.method_name(data[:method])
+    position = Contracts::ContractsSupport.method_position(data[:method])
+    method_name = Contracts::ContractsSupport.method_name(data[:method])
 
     header = if data[:return_value]
                "Contract violation for return value:"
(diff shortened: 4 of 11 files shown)


nixpulvis:
`Contracts::ContractsSupport` seems redundant. Why are you adding another support module to the `Contracts` namespace?

guizmaii:
It is redundant but it is now close to impossible to have any name collision.

I don't add any ```Support``` module to the ```Contracts``` namespace. I only have a Rails model named ```Support```. 
After put the Contract gem to my dependencies, some of my calls to my ```Support``` model are made with the ```Contracts::Support``` module and thus fail.

To be more precise, the problem comes after adding 2 gems to my dependencies : ```Contracts```  of course and ```grape``` (https://github.com/intridea/grape).

Whitout ```grape``` I don't have this problem. 

Maybe someone could reproduce my problem easily !

nixpulvis:
I see, the issue is because included with `Contracts` and thus get it's namespaces inside the classes you're working with. Frankly I've never been a fan of including contracts on every class for this very reason, however changing the name of the module doesn't seem like the right fix for the overarching problem. I'd be curious if we could either stop including contracts, or *only* import exactly what we need into the namespace.

guizmaii:
I've created an issue in the ```grape``` repo because the problem not only come from ```Contracts``` : https://github.com/intridea/grape/issues/1019

What do you mean by "including contracts on every class" ? I've included Contracts only on some of my classes and modules ! (Not the time to add Contracts on every methods for now unfortunatly)

The goal of this gem is to type (all) methods and methods parameters. If I can only type some of my methods, the interest of this gem is lower !

nixpulvis:
No sorry that's not what I'm saying. You can defiantly attach contracts to only the methods you want. However currently you do so by including `Contracts` into the namespace where you want to define the contracts, this leads to the potential for collisions.

I'm simply saying there could be room for this gem to be a bit more conservative with what it provides when you include it.

waterlink:
This is true the issue. Because by doing `include Contracts` you basically included everything that is namespaced under `Contracts::`, such as: `Support`, `Eigenclass`, `Engine`, etc/etc/etc.

Renaming `Support` => `ContractsSupport` will not fix the issue with others and it sounds strange: `Contracts::ContractsSupport` ;)

The real solution would be change `include Contracts` to `include Contracts::Core` or `Contracts::Engine` or `Contracts::Helpers` or something.

nixpulvis:
@waterlink yea that's what I was thinking. And then be very careful about the builtin contracts, or add a short namespace for all of them as well.

waterlink:
I think this will do:

```ruby
# enables the whole package: `Contract`, `Invariant` helpers and all builtin contracts
include Contracts::Full

# enables only `Contract`
include Contracts::Core

# enables only `Invariant`
include Contracts::Invariants

# enables unnamespaced builtin contracts
include Contracts::Builtin
```

Obviously, when you don't include builtin contracts, you will just reach them by `Contracts::Builtin::<NAME>`

Basically `Full` will just include all other three (or more in the future).

One thing is I am on the fence about `Full` - doesn't sound right to me. Any ideas?

WDYT?

waterlink:
*If anybody just got random notification, I pressed cmd+enter to early, sorry*

nixpulvis:
What about putting the builtins within the `Contracts` namespace so it's a little less verbose. `Contracts::Num` actually makes a lot of sense.

waterlink:
Then, how will you include them unnamespaced?

Actually I see a solution to that:

```ruby
module Contracts
  module Builtin
    class Nat
      # ...
    end

    # ...
  end

  Nat = Builtin::Nat
  # ...
end
```

+ a bit of metaprogramming to not duplicate this knowledge (iterate over direct constants in `Contracts::Builtin` namespace and alias them from `Contracts::` namespace)

nixpulvis:
Or just `include Builtin` in `Contracts`.

waterlink:
Ah, my bad, you are right :)

egonSchiele:
How about putting modules like `Support` under `Contracts::Internal` namespace?

waterlink:
That will not solve the problem. Because now you will clash with somebody's `Internal` domain object.

nixpulvis:
It's generally bad practice to have users `include` a namespace that is not explicitly designed to be included, especially the top level namespace. By changing the included module to something like `Core` we can have 100% control over what we add and what we don't.

waterlink:
@egonSchiele so, I assume, you just don't like to have `include Contracts::<ANYTHING>` and prefer just `include Contracts`

In that case we can re-namespace every tiny bit of the library to different namespace (something like `ContractsRuby` or `ContractsEngine`, or `ContractsInternals`) and make `Contracts = ContractsWhatever::Extension`

waterlink:
In that case:

- we are still able to do `include Contracts`, so hopefully no old code broken,
- some code is fixed (like the code of OP),
- we additionally give an ability to do more fine-grained inclusion in the new code, like: `include ContractsEngine::Core` and `include ContractsEngine::Invariants` and `include ContractsEngine::Builtin`, where `include Contracts` will just include all of them.

WDYT?

nixpulvis:
Of that should be easy. No need to muddy up a million names. Just override the `.append_features` hook for `Contracts` to only include what we want.

the namespaces should "make sense".

waterlink:
Yes, this is a good one. And good god it is supported by all the ruby versions in our travis :)

waterlink:
Actually, it looks like we want this hook to do nothing :) and in other 2 hooks (`included` and `extended`) we want to just to the usual stuff + include `Contracts::Builtin`. Are we?

waterlink:
Were able to repro this issue only with autoloading - actually when contracts and the class, that both includes contracts and uses conflicting name in its body comes before the class that defines the real user class. In this case user class doesn't get loaded at all.

guizmaii:
Happy to see how you are reactive ! :)

Your solution seams great. I'll test your PR this afternoon and give you feedback in the PR ;)

Thanks

waterlink:
This is better to be fixed by whatever comes out of #157 discussion.

PR #155, opened 2015-05-26
fixes #154: Don't include contracts internals into users classes

waterlink:
Looks like a good fix for #154. Thanks to @nixpulvis for the idea on `.append_features` hook.

Lets see if travis passes.

/cc @guizmaii : can you try this branch if it works for your application?

/ccc @egonSchiele @nixpulvis : WDYT on this solution?

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -17,6 +17,10 @@ module Contracts
     common(base)
   end
 
+  def self.append_features(base)
+    base.include(BuiltinContracts)
+  end
+
   def self.common(base)
     return if base.respond_to?(:Contract)
 
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -19,426 +19,438 @@ require "set"
 # The contract is <tt>Contract Num, Num, Num</tt>.
 # That says that the +add+ function takes two numbers and returns a number.
 module Contracts
-  # Check that an argument is +Numeric+.
-  class Num
-    def self.valid? val
-      val.is_a? Numeric
-    end
-  end
-
-  # Check that an argument is a positive number.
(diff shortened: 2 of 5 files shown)


waterlink:
BTW, since I wrapped all builtin contracts in their own namespace, you better read the PR with `?w=1` to hide all whitespace-only changes and see only relevant ones. I.e.: [just-use-this-link](https://github.com/egonSchiele/contracts.ruby/pull/155/files?w=1)

nixpulvis:
The approach looks good to me. I cannot attest to everyone that needs to be fixed because of it working.

guizmaii:
It doesn't seem to work. 

It seems to fix the ```Support``` name collision problem but other problems appear.

I have a ```Client``` model in my Rails app that has many ```Store```. With this patch, when I do something like that : ```client.stores``` it throw an exception : ```Client::Store constant does not exist```

Strange !

waterlink:
Can you pin it down to small example? Does this error go away if you rename
Client to something else? Does this error go away if you stop using grape
gem?

guizmaii:
I'll do some tests maybe tomorrow or next week. For now I don't have a lot of time.

nixpulvis:
For referance this is related to the larger issue of namespace collisions we should be weary of. Related to #157.

alex-fedorov:
@guizmaii Any news on this one? Have you been able to play around with it?

egonSchiele:
works for me!

alex-fedorov:
Hmm, I wonder. What happens in this case:

```ruby
module A
  include Contracts

  class B
  end
end

class X
  include A

  def b_class
    B
  end
end
```

waterlink:
I'm closing this for now, because #157 discussion may result in way better solution.

PR #156, opened 2015-05-29
fixes #150: generate Contracts class

### LINKED ISSUE #14: Integration with rspec/mocking
I'm not sure how contracts should integrate with rspec tests and mocking, especially stubbed things and weird testing calls where you try to send `nil` to a method that expects a `String`. Sometimes when I insert a contract, the rspec tests break, and I'm not sure whether this means I've discovered a new edge case, or whether Contracts is simply being incompatible with rspec.

For now, I'm treating rspec tests as sacrosanct and omitting contracts for some of my methods (e.g. initializers #11).

### LINKED ISSUE #150: Use code generation for inlining
The `Contract` class has been heavily inlined, with results in much of the code being closed for modification. I cannot subclass `Contract` and redefine how it handles Class contracts, for example, without rewriting an entirely new `make_validator` method.

Why not generate `contracts.rb` with ERB or other templating solution instead? That would allow the code to be broken down as it should be for maintainability, without sacrificing performance. (In particular, it would allow extending or monkeypatching `Contract` to play nicely with RSpec, rather than modifying RSpec itself, as in `contracts-rspec`.)

robnormal:
To generate the class statically, use `bin/generate-contracts`. This generates the "standard" Contracts class. To override code, you will have to use the admittedly awkward setup below. `Contracts.configure(:contracts_use_file => false)` causes the class to be generated on the fly and `eval`-ed, instead of loading from the existing file.

As an example, including the following at the top of `spec/spec_helper.rb` fixes #14:
```ruby
module Contracts
  @load_on_require = false
  require 'contracts'

  Contracts.configure(
    :contracts_generator => Contracts::ContractsGenerator.new(
      'Class' => "require 'rspec'; lambda { |arg| arg.is_a?(contract) || RSpec::Mocks::Double }"
    ),
    :contracts_use_file => false
  )
  Contracts.load
end
```
The template for `contract.rb` lives in `templates/contract.text.erb`. `ContractsGenerator` does the generating. So far, I've broken out only the part I needed to override - namely, the `lambda`s used to check Class and other basic contracts.


alex-fedorov:
@robnormal can you get rid of `require_relative` ? (travis failure) (`load` should be fine in this case)

@egonSchiele can you take a look at this one? I am still on the fence and have not decided on the API that we want to provide for such code inlining and overriding.

egonSchiele:
@robnormal thanks for taking the time to do this PR. I don't like the pre-compile step because the code is getting pretty complicated (I appreciate that you made a minimal amount of changes, but it's still very different from most ruby projects). I'm afraid this is going to make it hard for newcomers to understand the code. So it will slow down contribution and monkeypatching. I think I would rather find a traditional solution, even if it means adding a few more methods and marginally slowing things down.

alex-fedorov:
@robnormal I have an idea, how your problem can be fixed without code-generating, I will create a PR and will use your case as example soon.

PR #158, opened 2015-06-04
Update tutorial for #72

kliuless:


TUTORIAL.md
```diff
@@ -339,6 +339,12 @@ p map([1, 2, 3], lambda { |x| x + 1 }) # works
 p map([1, 2, 3], lambda { |x| "oops" }) # fails, the lambda returns a string.
 ```
 
+The above examples showed a method accepting a `Proc` as the last argument, but the same contract works on methods that accept a block:
```
waterlink: Wouldn't it be "..., but the same contract works on **the** methods ..."

kliuless: American English is my first language, and it's more proper to omit "the". Including "the" would mean you had actual methods in mind (i.e. you could enumerate them), while omitting "the" is used to describe a category of methods without being able to enumerate.
This is in different from some other languages, where you would use the definite article for categories.

waterlink: You are right these are not specific methods, but general category.

waterlink:
Thanks @kliuless.

PR #166, opened 2015-06-28
Tutorial: new section "Using contracts within your own code"

gsinclair:
A brief section near the end to describe the use of `Contract.valid?`.

Also fixed a single typo (separate commit).

Aside: I think the comment for `Contract.valid?` is out of date. It says it returns a tuple but I think it just returns true/false.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -652,6 +652,17 @@ If you run it, last line will generate invariant violation:
 
 Which means, that after `#silly_next_day!` all checks specified in `invariant` statement will be verified, and if at least one fail, then invariant violation error will be raised.
 
+## Using contracts within your own code
+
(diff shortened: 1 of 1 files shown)


waterlink:
Needs to be rebased against master

egonSchiele:
I don't think this is an introductory tutorial anymore -- it is a comprehensive doc. Maybe we should think about splitting this tutorial up into a quick "getting started" guide and a longer "full documentation".

gsinclair:
The "tutorial", whether it's a tutorial or not, is very well structured.
It's easy enough to read it right through and get a good sense of the
library, and then come back to refer to things as needed.

A "getting started" would be a good idea. I don't know what you'd have in
it, and I guess it would be very brief, but I _do_ think that even this
shorter guide should have the full list of built-in contracts. You don't
get a sense of how useful this library will be unless you see all of that.

PR #172, opened 2015-07-03
[rspec update] Fix rspec double related spec

waterlink:
Recent rspec release changed how double's inspect behaves slightly. So we need to adjust this one spec.


waterlink:
Will merge this when green

egonSchiele:
Thanks for fixing.

PR #173, opened 2015-07-03
Fixed #165: "Bug: comparison of Symbol with 0 failed"

### LINKED ISSUE #165: Bug: comparison of Symbol with 0 failed
To reproduce:

    Contract Nat => Any
    def foo(n) n+1 end
    foo 6                           # Works as expected
    foo -10                         # Gives the correct error message
    foo :x                          # Fails with ArgumentError instead of contract violation

gsinclair:
This bug was in Nat. Also fixed equivalent bugs in Pos and Neg.

Also beefed up a few tests for Pos, Neg, Nat.

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -29,21 +29,21 @@ module Contracts
   # Check that an argument is a positive number.
   class Pos
     def self.valid? val
-      val && val > 0
+      val && val.is_a?(Numeric) && val > 0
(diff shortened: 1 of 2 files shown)


waterlink:
Can you run `rubocop` locally and fix the issues it reports ?

gsinclair:
Done. (There are a lot remaining elsewhere in the code though.)

egonSchiele:
Thanks!

PR #177, opened 2015-07-08
Update changelog for 0.10

waterlink:
So this is the list of important changes between `0.9` and `0.10`. My wording is probably not the best, so please review :)

/cc @egonSchiele @PikachuEXE


PikachuEXE:
I think it is good enough, though having links to PRs would be even better :)

alex-fedorov:
@PikachuEXE Done

PikachuEXE:
Nice! :+1:

egonSchiele:
Awesome! Links are really nice :)

PR #180, opened 2015-07-16
Update bundler before doing bundle install

alex-fedorov:
Workaround for travis ruby 2.1 built-in bundler of being too old.

reference: https://github.com/travis-ci/travis-ci/issues/3531


alex-fedorov:
Will merge this if green, since it is no-op.

dg-ratiodata:
Green. :-)

PR #183, opened 2015-07-17
Add contract to check class AND sub-class

dg-ratiodata:
This PR adds a contract which checks the class and the sub-class of a given object.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,6 @@
+AllCops:
+  DisplayCopNames: true
+
 # forces method defs to have params in parens
 Style/MethodDefParentheses:
   Enabled: false
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -99,6 +99,7 @@ contracts.ruby comes with a lot of built-in contracts, including the following:
 
 * Miscellaneous
   * [`Exactly`](http://www.rubydoc.info/gems/contracts/Contracts/Exactly) – checks that the argument has the given type, not accepting sub-classes, e.g. `Exactly[Numeric]`.
+  * [`IsA`](http://www.rubydoc.info/gems/contracts/Contracts/IsA) – checks that the argument is of given class or sub-class of given class, e.g. `IsA[Numeric]`
   * [`Eq`](http://www.rubydoc.info/gems/contracts/Contracts/Eq) – checks that the argument is precisely equal to the given value, e.g. `Eq[String]` matches the class `String` and not a string instance.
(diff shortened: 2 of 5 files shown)


alex-fedorov:
This is not really needed, because this works already:

```ruby
class A; end
class B < A; end

Contract A => nil
def something(x) end

something(A.new)    # => nil
something(B.new)    # => nil
something(Object.new)   # raise ContractError
```

alex-fedorov:
It is implemented in validators: https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts/validators.rb#L44-L46

dg-ratiodata:
Cool. Didn't know that. Is there any documentation I oversaw?

waterlink:
Probably not. It is just how type restriction in typed languages naturally
work.

We might want to add this to documentation, but the problem is that we have
only tutorial, that is already quite bloated. We might as well want to
split it in two. WDYT @egonSchiele?

dg-ratiodata:
Why not using `cucumber`+ `aruba` for this? Although I'm a bit biased. :smile:

The guys at rspec use cucumber to describe how to use their software. `aruba`
is used behind the scenes. Might be worth to checkout for you as well. We at `aruba` like that idea and started to migrate our very own features to such a theme. They look something like that.

```cucumber
Feature: Checking Class Type
  Scenario: Object is of Class
    Given a file named "lib/my_lib/my_class.rb" with
    """
    require 'contracts'

    module MyLib
      class MyClass
        include Contracts

        Contract Numeric => nil
        def do_some_work(arg)
        end
      end
    end
    """
    And a file named "spec/class_type_spec.rb" with:
    """
    require 'spec_helper'
    require_relative 'my_lib/my_class' # just to demonstrate, normally I would move that thing into the spec helper

    RSpec.describe "Class Type Checking" do
      subject { MyClass.new }

      context 'when is valid' do
        let(:arg) { 1 }
        it do
          expect do
            subject.do_some_work
          end.not_to raise_error
        end
      end

      context 'when is invalid' do
        let(:arg) { 'asdf' }
        it do
          expect do
            subject.do_some_work
          end.to raise_error ArgumentError
        end
      end
    end
    """
    When I run `rspec`
    Then the specs should all pass
```

The good thing is, you've got tests you can run, but also a pretty good documentation your users can use. Maybe you could even upload that stuff to ["relish"](https://www.relishapp.com/) if you like.

alex-fedorov:
@dg-ratiodata Just check out this: https://github.com/egonSchiele/contracts.ruby/tree/0.99-before-release :D

alex-fedorov:
This is old non-updated branch, but soon we will take care of it in preparation of `1.0.0` and it already has cucumber/aruba scenarios in it :)

dg-ratiodata:
Great!

egonSchiele:
Yes, I was just thinking about that this week. I can split out the tutorial.

PR #184, opened 2015-07-22
Add Range validator

### LINKED ISSUE #167: Add range as a contract
To say that this function expects a number between 1 - 10. Options:

```ruby
Contract (1..10) => Bool
```

OR

```ruby
Contract InRange[1..10] => Bool
```

We might have to reserve `1..10` (from the first example) so people can write contracts for functions that expect a range as an argument. In which case we would implement the second solution.

waterlink:
fixes #167 

Implemented first option:

```ruby
Contract (1..10) => String
def do_something(range)
```

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -546,6 +546,7 @@ Possible validator overrides:
 - `override_validator(Proc)` - e.g. `lambda { true }`,
 - `override_validator(Array)` - e.g. `[Num, String]`,
 - `override_validator(Hash)` - e.g. `{ :a => Num, :b => String }`,
+- `override_validator(Range)` - e.g. `(1..10)`,
 - `override_validator(Contracts::Args)` - e.g. `Args[Num]`,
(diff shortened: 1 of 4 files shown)


gsinclair:
Nice. What happens if you pass a string, say, to a 1..10 argument? And what should happen?

alex-fedorov:
Good question. Will add spec for that ;)

alex-fedorov:
Well, as I think about it further, what should happen? One of:

- `ContractError: Expected: 1..10; Actual: "hello world"` - this is how it works now;
- `ContractError: Expected: Range(Numeric); Actual: String` - this is nice, but what to do if you have a range of some weird type that has inheritance and stuff like this, how to determine base type?

WDYT?

gsinclair:
I think how it works now is fine. Worth making a spec though.

waterlink:
@gsinclair I actually did added spec for this before adding my last comment ;)

egonSchiele:
thanks!

abevoelker:
Ooh I was just looking for a Range contract last night - nice timing and thank you!

Issue #157, opened 2015-06-01
Do not require importing contract class names into client classes


robnormal:
On a large project, there may be a large number of contracts you will want to name, and these names will then be included in every class that includes the `Contracts` module. The potential for naming collisions could grow very quickly. (I am already having this problem - I have a class called "Maybe".)

One possible fix would be an alternative include, in which the `Contract` method takes a block instead of a hash.
```ruby
class Alpha
  include ContractsSafe
  Contract { ArrayOf[Num] => Maybe[Or[This, That]] }
  # ...
end
```
The block could then be `instance_eval`-ed elsewhere, isolating the contract names.

nixpulvis:
Related #101 and #105.

alex-fedorov:
This code is not a block now, but normal hash, probably you want something:

```
Contract { [ ArrayOf[Num] => Maybe[Or[This, That]] ] }
```

robnormal:
Actually, it's a syntax error! The equivalent would be
```ruby
Contract { { ArrayOf[Num] => Maybe[Or[This, That]] } }
```
which I have to admit is pretty ugly. Another possibility entirely - which would solve some other issues, including the load-order problem - would be to use a string to specify the contract. This could be `eval`-ed, or, better yet, parsed. I realize this is a pretty different direction than the project is currently going in, though.

waterlink:
Hm, not really:

```
2.2.1 :001 > def Contract(*, &blk)
2.2.1 :002?>   end
 => :Contract 
2.2.1 :003 > Contract { [ String => String ] }
 => nil 
2.2.1 :004 >
```

waterlink:
You are wrong here, it is still possible to shape this project if it is worth it. We had long conversation on gitter some time ago about this and arrived at conclusion that we want to have block/eval/whatever there, because it allows to have more advanced type system.

nixpulvis:
The trick is going to be moving forward with this while maintaining functionality.

waterlink:
Yeah, of course =) Not really a big of a problem. And remember, we can always break an API, especially if we find some better API than current one (I doubt so, but everything is possible), if we provide nice deprecation-removal curve.

waterlink:
Supporting current API and block-based one, not really a big deal, since it is totally different calls and may even result in instantiating totally different decorator classes for instance.

nixpulvis:
[...]

PR #185, opened 2015-07-22
Refactor :: Extract contracts core

waterlink:
This is a first stab on #157.

This refactoring does:

- Extracts `Contracts::Core` out of `Contracts`;
- Makes `include Contracts` include `Contracts::Core` for backwards compatibility;
- Factors out specs to use `Contracts::Core`;
- Makes `TUTORIAL.md` use `Contracts::Core` use it by default to encourage new users of contracts to use `include Contracts::Core` right away;
- Mentions shortcut `C = Contracts` in `TUTORIAL.md`.

/cc @egonSchiele @sfcgeorge @nixpulvis @robnormal

### DIFF
--- a/spec/fixtures/fixtures.rb
+++ b/spec/fixtures/fixtures.rb
@@ -1,9 +1,11 @@
 require "date"
 
-class A
-  include Contracts
+C = Contracts
 
-  Contract Num => Num
+class A
+  include Contracts::Core
+
+  Contract C::Num => C::Num
   def self.a_class_method x
     x + 1
   end
@@ -12,12 +14,12 @@ class A
     true
   end
 
-  Contract Num => Num
+  Contract C::Num => C::Num
   def triple x
     x * 3
   end
 
-  Contract Num => Num
+  Contract C::Num => C::Num
   def instance_and_class_method x
     x * 2
   end
@@ -29,7 +31,7 @@ class A
 end
 
 class B
-  include Contracts
+  include Contracts::Core
 
   def bad
     false
@@ -41,8 +43,8 @@ class B
   end
 end
 
-class C
-  include Contracts
+class F
+  include Contracts::Core
 
   def good
     false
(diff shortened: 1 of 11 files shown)


gsinclair:
Just my point of view, I guess, but wouldn't "include Contracts" be the sensible default usage? Fair enough that sometimes you need to be careful to avoid name clashes, and therefore "include Contracts::Core", but wouldn't that be a minority of cases?

waterlink:
Not really. `include Contracts` already includes to much. For example `Support`, `Decorator`, `Formatters` - those are quite common names.

waterlink:
We want to deprecate bare `include Contracts` because of this. (deprecation for `0.99.x` and removal for `1.0.0`)

waterlink:
You will be still able to include builtin contracts with `include Contracts::Builtin` or something, but it is out of scope of this PR and will be done probably for `1.0.0`.

spec/fixtures/fixtures.rb
```diff
@@ -506,9 +508,9 @@ def next_year!
 
 class SingletonClassExample
   # This turned out to be required line here to make singleton classes
-  # work properly under all platforms. Not sure if it worth trying to
+  # work properly under all platforms. C::Not sure if it worth trying to
```
waterlink: Need to remove this change :D

spec/fixtures/fixtures.rb
```diff
@@ -274,31 +276,31 @@ def tutorial_map(arr, func)
     ret
   end
 
-  # Need to test Func with weak contracts for other args
+  # Need to test C::Func with weak contracts for other args
```
waterlink: Same here

waterlink:
Rebased against upstream master to be mergeable

egonSchiele:
Wow, nice work! Really excited about this.

alex-fedorov:
Rebased again. Such branches is hard to maintain, since they touch almost all the testing code and each merged PR makes it non-merge-able .

@egonSchiele Can you merge it? or do you have any concerns?

sfcgeorge:
I think it's nice, a good step to making Contracts more compatible. I like that the old `include Contracts` works the same as the new syntax is more verbose and unnecessary for small projects where clashes are unlikely. I'm hoping that in the future bracket box syntax `Contracts {[ ... ]}` could be added for terseness and no collisions.

But I think this could be merged as it adds clash free for those who need it and is still backwards compatible.

egonSchiele:
Merged!

PR #187, opened 2015-07-22
Make KeywordArgs fail if unexpected keys are passed in

abevoelker:
Wasn't sure where to put the KeywordArgs specs - let me know if it should go in spec/contracts_spec.rb or whatnot (or if I'm otherwise way off base)

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -385,6 +385,7 @@ module Contracts
     end
 
     def valid?(hash)
+      return false unless hash.keys - options.keys == []
       options.all? do |key, contract|
         Optional._valid?(hash, key, contract)
       end
--- a/spec/builtin_contracts_spec.rb
+++ b/spec/builtin_contracts_spec.rb
@@ -332,6 +332,24 @@ RSpec.describe "Contracts:" do
     end
   end
 
+  describe "KeywordArgs:" do
+    it "should pass for exact correct input" do
+      expect { @o.person_keywordargs(:name => "calvin", :age => 10) }.to_not raise_error
+    end
+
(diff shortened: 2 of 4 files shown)


abevoelker:
Hmm weird, specs passed with 2.2 on my machine. I'll look into the failures.

abevoelker:
Ah `bundle exec rake` was passing, but it seems that `bundle exec rspec` does fail on my machine as it runs more specs than rake...

alex-fedorov:
Looks like you actually want to fix this spec. Because it tries to pass in unexpected argument and expects it to not fail. Seems like a buggy behavior to me.

abevoelker:
Okay I think I fixed it - let me know if I violated the intent of the spec or whatnot

lib/contracts/builtin_contracts.rb
```diff
@@ -385,9 +385,10 @@ def initialize(options)
     end
 
     def valid?(hash)
-      options.all? do |key, contract|
-        Optional._valid?(hash, key, contract)
-      end
+      (hash.keys - options.keys == []) &&
```
alex-fedorov: Can you use guard statement here instead - current code makes it hard to read. Example:

```ruby
def valid?(hash)
  return false unless hash.keys - options.keys == []      # <--- this is a guard statement
  # .. do the main job here ..
end
```

alex-fedorov:
Other than that, looks good.

lib/contracts/builtin_contracts.rb
abevoelker: Oh yeah for sure, should've asked about that

alex-fedorov:
@abevoelker Thanks!

Issue #161, opened 2015-06-10
Varargs with a block - Stack overflow


alem0lars:
Combining varargs with a block leads to stack overflow, because you always assume that you will give a proc.

See https://github.com/middleman/middleman/issues/1494

waterlink:
Trying to reproduce:

```
2.2.1 :009 > class X
2.2.1 :010?>   include Contracts
2.2.1 :011?>   Contract Any, Args[Any], Maybe[Proc] => Any
2.2.1 :012?>   def use(middleware, *args, &block)
2.2.1 :013?>     p [middleware, args, block].inspect
2.2.1 :014?>     :a_result
2.2.1 :015?>     end
2.2.1 :016?>   end
 => :use 
2.2.1 :017 > x = X.new
 => #<X:0x000000023bfc98> 
2.2.1 :018 > x.use(:a_middleware, :bga, :test) { :helloworld }
"[:a_middleware, [:bga, :test], #<Proc:0x000000023a6450@(irb):18>]"
 => :a_result 
2.2.1 :019 > x.use(:a_middleware, :bga, :test)
"[:a_middleware, [:bga, :test, nil], nil]"
 => :a_result
```

Don't get stack overflow, but I see, that in case block is nil, we get additional argument in `args` splat. Which can damage logic and cause any problems. In case of `middleman` it caused stack overflow, but can cause any logic problems.

Needs to be fixed..

egonSchiele:
Ok, Since @alem0lars commented in the middleman issue that the stack overflow isn't happening anymore, I'm closing this issue as fixed.

PR #188, opened 2015-07-23
bug fix for varargs with a maybe[block]

egonSchiele:
When varargs are given with a maybe[block], and the block is not provided, do not append an extra nil to the varargs (see waterlink's issue #161).

What do you think of this @waterlink ?


waterlink:
LGTM

PR #190, opened 2015-07-25
Do not allow non-hash values in HashOf contract

abevoelker:
Given these HashOf contracts:

```ruby
Contract HashOf[Symbol => String] => Any
def foo(args)
end

Contract KeywordArgs[:hash => HashOf[Symbol => String]] => Any
def bar(args)
end
```

calling `foo(nil)` or `bar(hash: nil)` results in `NoMethodError: undefined method ``keys' for nil:NilClass`

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -360,6 +360,7 @@ module Contracts
     end
 
     def valid?(hash)
+      return false unless hash.is_a?(Hash)
       keys_match = hash.keys.map { |k| Contract.valid?(k, @key) }.all?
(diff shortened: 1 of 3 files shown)


abevoelker:
Actually I amended the commit since it wasn't really a thorough fix - you could pass in pretty much anything that isn't a Hash to trigger the error, such as `foo(1)`. I don't know if `is_a?(Hash)` is too strict of a check or whatnot so let me know if there is a better way to do it.

waterlink:
LGTM

waterlink:
@abevoelker Good catch! Thanks!

PR #196, opened 2015-08-30
Add match-contract

eval:
For a recent project this contract would have been handy:

```ruby
RGB = Match[/\#h{6}/] # e.g. '#ffffff'

Contract RGB => nil
def assign_color(color)
  # 
end
```

Other example:
```ruby
Contract Match[default_mail_regexp] => nil
def send_invitation(email)
...
```

As it's not easy to get the effect of a logical AND with regular expressions, so I added that as well:

```ruby
PartnerEmail = Match[default_mail_regexp, /(company1|company2)\.com\z/]

Contract PartnerEmail => nil
def invite_to_project(email)
```

Hopefully it's useful to others! If so I'm happy to move the explanation to the tutorial.

### DIFF
--- a/spec/validators_spec.rb
+++ b/spec/validators_spec.rb
@@ -1,9 +1,9 @@
 require "spec_helper"
 
-module Contracts
-  RSpec.describe "range contract validator" do
-    subject(:o) { GenericExample.new }
+RSpec.describe "Contract validators" do
+  subject(:o) { GenericExample.new }
 
+  describe "Range" do
     it "passes when value is in range" do
       expect do
         o.method_with_range_contract(5)
@@ -20,6 +20,28 @@ module Contracts
       expect do
         o.method_with_range_contract("hello world")
       end.to raise_error(ContractError, /Expected: 1\.\.10.*Actual: "hello world"/m)
+    end
+  end
+
+  describe "Regexp" do
+    it "should pass for a matching string" do
+      expect { o.should_contain_foo("containing foo") }.to_not raise_error
+    end
+
+    it "should fail for a non-matching string" do
+      expect { o.should_contain_foo("that's not F00") }.to raise_error(ContractError)
+    end
+
+    describe "within a hash" do
+      it "should pass for a matching string" do
+        expect { o.hash_containing_foo(:host => "foo.example.org") }.to_not raise_error
+      end
+    end
+
+    describe "within an array" do
+      it "should pass for a matching string" do
+        expect { o.array_containing_foo(["foo"]) }.to_not raise_error
+      end
     end
   end
 end
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -271,6 +271,22 @@ give_largest_value(a: 1, b: 2, c: 3) # returns 3
 give_largest_value("a" => 1, 2 => 2, c: 3)
 ```
 
+### Contracts On Strings
+
+When you want a contract to match not just any string (i.e. `Contract String => nil`), you can use regular expressions:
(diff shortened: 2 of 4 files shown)


nixpulvis:
Could we use the bare regex, for example:

```ruby
Contract /\#h{6}/ => nil
```

waterlink:
According to:
https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts/validators.rb#L3-L57
- it is not possible to match with bare regex currently. In case we find it
as an appropriate way to match regex, one could just add new key with
corresponding lamda there.

--
Best Regards,
Alexey Fedorov,
+49 15757 486 476

2015-08-31 2:05 GMT+02:00 Nathan Lilienthal <notifications@github.com>:

egonSchiele:
I like this contract! It would be a lot more elegant as a bare regex as @nixpulvis suggested. You could still get the AND functionality by using the And contract: `And[/some regex/, /another regex/]`. Do you feel comfortable adding your contract to this hash, @eval ?

https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts/validators.rb#L3-L57

eval:
Much cleaner indeed, will update. Thanks for the pointers!

eval:
I updated the code and added some explanation to `TUTORIAL.md` /cc @egonSchiele 

Note that the current version cannot be used in something like
```ruby
Contract (a: C::Match[/foo/]) => nil
```
Should we support this as well?

waterlink:
Hm, I think validators should work correctly inside of hashes. So this should work:

```ruby
Contract {a: /foo/} => nil
def bar(hash)
end
```

spec/builtin_contracts_spec.rb
```diff
@@ -360,6 +360,16 @@
     end
   end
 
+  describe "Regexp:" do
```
waterlink: Can you move these specs to `validators_spec`?

waterlink:
Yeah, hash validator here treats values as contracts too, so it is possible to have them as a regexp and regexp validator will kick in for them. Here: https://github.com/eval/contracts.ruby/blob/feature/match-contract/lib/contracts/validators.rb#L19-L26

eval:
@waterlink this works indeed within other contracts :+1:

I moved the specs. Let me know if there's anything else.

spec/validators_spec.rb
```diff
@@ -1,9 +1,9 @@
 require "spec_helper"
 
-module Contracts
-  RSpec.describe "range contract validator" do
-    subject(:o) { GenericExample.new }
+RSpec.describe 'Contract validators' do
```
waterlink: rubocop says here to switch to double quote string.

spec/validators_spec.rb
```diff
@@ -22,4 +22,26 @@ module Contracts
...
+      expect { o.should_contain_foo("containing foo") }.to_not raise_error
+    end
+
+    it "should fail for a non-matching string" do
+      expect { o.should_contain_foo("that's not F00") }.to raise_error(ContractError)
+    end
+
+    describe 'within a hash' do
+      it "should pass for a matching string" do
+        expect { o.hash_containing_foo(host: "foo.example.org") }.to_not raise_error
```
waterlink: Can you use here `1.8` hash rocket syntax, since we still want to support `1.8` ?

eval:
Amended the style-fixes - thanks for your feedback @waterlink

waterlink:
LGTM

waterlink:
Thanks @eval Good work!

PR #198, opened 2015-09-09
Fix classes that extend modules where contracts were included

### LINKED ISSUE #176: Cannot call Contract in AS#included block
# Env
MRI `2.2.2` or `2.2.3`

# What happened
- I extended a class (e.g. `Post`) with a module (e.g. `ClassMethods`) where `ClassMethods` has included `Contracts`
- When I call `Contract` method inside `Post` after including `ClassMethods`:
  - It works fine in `0.9`
  - It raises an error in `0.10.1` or `0.11.0` (see the `Error` section below)

To reproduce the error please look at the [test script gist](https://gist.github.com/PikachuEXE/e7efb3a58ef3c9764722)
Too long to read? Shorter version: https://gist.github.com/PikachuEXE/1672d5c0434863bee6ae

# Error
Message from application: undefined method `Contract' for Class:Class (NoMethodError)

# Updates
**Update 1**: Even after I have removed `ActiveSupport::Concern` usage (and switch to `Module#included`, it's still failing. So I updated the code to ensure it's not AS related.
**Update 2**: Included Ruby versions
**Update 3**: Added [test script gist](https://gist.github.com/PikachuEXE/e7efb3a58ef3c9764722)
**Update 4**: Update the test script again (removed AS dependency)

waterlink:
Fixes #176 

Our current routine while including or extending `Contracts` is to define `Contract(*args)` method for current class and make it just delegate to `self.class.Contract`. But in case of:

```ruby
m = Module.new do
  include Contracts::Core
end

Class.new do
  include Contracts::Core
  extend m

  Contract String => nil
  def foo(x)
  end
end
```

it doesn't work, since there is no `self.class.Contract` at this point of time. But, there is just `Contract` already in super class, and we may just call it. That is what this PR does: verifies if there is a `Contract` method in super class and calls it instead of delegating to `self.class.Contract`.

/cc @PikachuEXE @egonSchiele

### DIFF
--- a/lib/contracts/core.rb
+++ b/lib/contracts/core.rb
@@ -24,7 +24,10 @@ module Contracts
         end
       end
 
-      base.class_eval do
+      # NOTE: Workaround for `defined?(super)` bug in ruby 1.9.2
+      # source: http://stackoverflow.com/a/11181685
+      # bug: https://bugs.ruby-lang.org/issues/6644
+      base.class_eval <<-RUBY
         # TODO: deprecate
         # Required when contracts are included in global scope
         def Contract(*args)
@@ -34,7 +37,9 @@ module Contracts
             self.class.Contract(*args)
           end
         end
+      RUBY
 
+      base.class_eval do
         def functype(funcname)
           contracts = Engine.fetch_from(self.class).decorated_methods_for(:instance_methods, funcname)
           if contracts.nil?
--- a/lib/contracts/core.rb
+++ b/lib/contracts/core.rb
@@ -28,7 +28,11 @@ module Contracts
         # TODO: deprecate
         # Required when contracts are included in global scope
         def Contract(*args)
-          self.class.Contract(*args)
+          if defined?(super)
+            super
+          else
+            self.class.Contract(*args)
+          end
         end
 
         def functype(funcname)
--- a/spec/contracts_spec.rb
+++ b/spec/contracts_spec.rb
@@ -717,4 +717,33 @@ RSpec.describe "Contracts:" do
       expect { c.double("asd") }.to raise_error
     end
   end
+
+  describe "classes with extended modules" do
+    let(:klass) do
+      m = Module.new do
+        include Contracts::Core
+      end
+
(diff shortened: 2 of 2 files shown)


waterlink:
@PikachuEXE Can you verify that this branch https://github.com/waterlink/contracts.ruby/tree/fix/176/classes_with_extended_modules_with_contracts fixes the problem for you?

waterlink:
Hmm, this fix doesn't work with ruby 1.9.2 only.

waterlink:
Ok, now it works for every ruby :)

PikachuEXE:
@waterlink 
[Your branch](https://github.com/waterlink/contracts.ruby/tree/fix/176/classes_with_extended_modules_with_contracts) works for the test script **and** my project :)
Which is #198 ?

PikachuEXE:
Wait I am in #198 now

PikachuEXE:
I got another question
Should I include `Contracts::Core` or `Contracts`? (for latest versions)
I don't want to use the deprecated one :)

alex-fedorov:
`Contracts::Core` is better in terms of not having all our helper classes in your namespace. But it means you don't have builtin contracts either.

Before we fully forbid usage of `include Contracts`, we will add `include Contracts::Builtin`. This way, you will be able to do:

```ruby
include Contracts::Core
include Contracts::Builtin
```

And you will get original functionality that you have had before with `include Contracts` but without our helper classes in your namespace.

egonSchiele:
lgtm

PikachuEXE:
Hope to see `Contracts::Builtin` soon :)

alex-fedorov:
@PikachuEXE this is very simple change. Maybe you are up for making a PR?

PikachuEXE:
`Contracts::Builtin` or `Contracts::BuiltinContracts`?

alex-fedorov:
`Contracts::BuiltinContracts` sounds like a repetition (you already have `Contracts::` at beginning), so `Contracts::Builtin` it is.

PR #199, opened 2015-09-14
Refactor/extract builtin contracts

PikachuEXE:
This is for [a comment in another PR](https://github.com/egonSchiele/contracts.ruby/pull/198#issuecomment-139492852)

# Changes
- Create module `Builtin` as suggested
- Change all references in code
- Change all references in doc
- Auto include `Builtin` when module `Contracts` included 
  (but not when extended, since I am not sure)

# Missing stuff
- Test for built-in contract usage without namespacing when `Contracts` included

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -108,7 +108,7 @@ class Contract < Contracts::Decorator
   def to_s
     args = args_contracts.map { |c| pretty_contract(c) }.join(", ")
     ret = pretty_contract(ret_contract)
-    ("#{args} => #{ret}").gsub("Contracts::", "")
+    ("#{args} => #{ret}").gsub("Contracts::Builtin::", "")
   end
 
   # Given a hash, prints out a failure message.
--- a/lib/contracts/formatters.rb
+++ b/lib/contracts/formatters.rb
@@ -103,7 +103,7 @@ module Contracts
       end
 
       def strip_prefix(val)
-        val.gsub(/^Contracts::/, "")
+        val.gsub(/^Contracts::Builtin::/, "")
       end
     end
 
--- a/lib/contracts/formatters.rb
+++ b/lib/contracts/formatters.rb
@@ -90,7 +90,7 @@ module Contracts
 
       def plain?
         # Not a type of contract that can have a custom to_s defined
-        !@value.is_a?(CallableClass) && @value.class != Class
+        !@value.is_a?(Builtin::CallableClass) && @value.class != Class
       end
 
       def useful_to_s?
--- a/spec/builtin_contracts_spec.rb
+++ b/spec/builtin_contracts_spec.rb
@@ -392,7 +392,7 @@ RSpec.describe "Contracts:" do
       end
 
       context "given String => Num" do
-        it { expect(Contracts::HashOf[String, Contracts::Num].to_s).to eq("Hash<String, Contracts::Num>") }
+        it { expect(Contracts::HashOf[String, Contracts::Num].to_s).to eq("Hash<String, Contracts::Builtin::Num>") }
       end
     end
   end
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -19,460 +19,465 @@ require "set"
 # The contract is <tt>Contract Num, Num, Num</tt>.
 # That says that the +add+ function takes two numbers and returns a number.
 module Contracts
-  # Check that an argument is +Numeric+.
-  class Num
-    def self.valid? val
-      val.is_a? Numeric
-    end
-  end
-
-  # Check that an argument is a positive number.
-  class Pos
-    def self.valid? val
-      val && val.is_a?(Numeric) && val > 0
-    end
-  end
-
-  # Check that an argument is a negative number.
-  class Neg
-    def self.valid? val
-      val && val.is_a?(Numeric) && val < 0
-    end
-  end
-
-  # Check that an argument is a natural number.
-  class Nat
-    def self.valid? val
-      val && val.is_a?(Integer) && val >= 0
-    end
-  end
-
-  # Passes for any argument.
-  class Any
-    def self.valid? val
-      true
-    end
-  end
-
-  # Fails for any argument.
-  class None
-    def self.valid? val
-      false
-    end
-  end
-
-  # Use this when you are writing your own contract classes.
-  # Allows your contract to be called with <tt>[]</tt> instead of <tt>.new</tt>:
-  #
-  # Old: <tt>Or.new(param1, param2)</tt>
-  #
-  # New: <tt>Or[param1, param2]</tt>
-  #
-  # Of course, <tt>.new</tt> still works.
-  class CallableClass
-    include ::Contracts::Formatters
-    def self.[](*vals)
-      new(*vals)
-    end
-  end
-
-  # Takes a variable number of contracts.
-  # The contract passes if any of the contracts pass.
-  # Example: <tt>Or[Fixnum, Float]</tt>
-  class Or < CallableClass
-    def initialize(*vals)
-      @vals = vals
-    end
-
-    def valid?(val)
-      @vals.any? do |contract|
-        res, _ = Contract.valid?(val, contract)
-        res
+  module Builtin
+    # Check that an argument is +Numeric+.
+    class Num
+      def self.valid? val
+        val.is_a? Numeric
       end
     end
 
-    def to_s
-      @vals[0, @vals.size-1].map do |x|
-        InspectWrapper.create(x)
-      end.join(", ") + " or " + InspectWrapper.create(@vals[-1]).to_s
-    end
-  end
-
-  # Takes a variable number of contracts.
-  # The contract passes if exactly one of those contracts pass.
-  # Example: <tt>Xor[Fixnum, Float]</tt>
-  class Xor < CallableClass
-    def initialize(*vals)
-      @vals = vals
-    end
-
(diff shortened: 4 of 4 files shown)


alex-fedorov:
You can include Builtin into Contracts directly at a module level. Then you
don't need to call include Builtin inside of included.

This way you can still refer them as Contracts::Maybe anywhere.

Then I would add to tutorial example what happens if including
Contracts::Builtin directly

PikachuEXE:
So the desired objectives are the following?:
1. Be able to use built in contracts without namespace after `Contracts` included
2. Be able to include `Contracts::Core` & `Contracts::Builtin`, and still works like (1)
3. Be to reference built in contracts as something like `Contracts::Maybe`

alex-fedorov:
@PikachuEXE Exactly!

PikachuEXE:
Updated & squashed

In the next major bump (breaking),
Is that only `Contracts::Builtin::Maybe` is usable (or preferred)?

waterlink:
No, we want to always have `Builtin` included in `Contracts`, so `Contracts::Maybe` is usable and preferred.

Extraction to `Builtin` is only done to make it possible to `include Contracts::Builtin` along with `Contracts::Core` when your code base don't have name clashes with built-in contracts.

Just `include Contracts` is a no go, because we define much more stuff than just built-in contracts, like: `Engine` and `Support` - which are pretty likely to clash.

TUTORIAL.md
```diff
@@ -17,7 +17,7 @@ contracts.ruby brings code contracts to the Ruby language. Code contracts allow
 A simple example:
 
 ```ruby
-Contract Contracts::Num, Contracts::Num => Contracts::Num
+Contract Contracts::Builtin::Num, Contracts::Builtin::Num => Contracts::Builtin::Num
```
waterlink: I think, we don't want to expose this information to users. We want them to still either:
- include `Contracts::Core` and use `Contracts::Maybe` (or `C::Maybe`)
- include `Contracts::Core`, `Contracts::Builtin` and use `Maybe`

So in `TUTORIAL.md` good changes are only for error messages and `validator` list

features/builtin_contracts/README.md
```diff
@@ -1,7 +1,7 @@
 To use builtin contracts you can refer them with `Contracts::*`:
 
 ```ruby
-Contract Contracts::Num => Contracts::Maybe(Contracts::Num)
+Contract Contracts::Builtin::Num => Contracts::Builtin::Maybe(Contracts::Builtin::Num)
```
waterlink: This does not need to change.

lib/contracts.rb
```diff
@@ -12,6 +12,11 @@
 require "contracts/core"
 
 module Contracts
+  # For backward compatibility
```
waterlink: Actually, this is a shortcut for users who don't want to `include Contracts::Builtin` (because of nameclashes), so that they can type `Contracts::Maybe` or `C::Maybe` (if they have aliased it)

lib/contracts.rb
```diff
@@ -72,7 +77,7 @@ def initialize(klass, method, *contracts)
     end
 
     @args_contract_index = args_contracts.index do |contract|
-      contract.is_a? Contracts::Args
+      contract.is_a? Contracts::Builtin::Args
```
waterlink: I think our code is fine using just `Contracts::Args` here and at all other places.

lib/contracts.rb
```diff
@@ -108,7 +113,7 @@ def pretty_contract c
   def to_s
     args = args_contracts.map { |c| pretty_contract(c) }.join(", ")
     ret = pretty_contract(ret_contract)
-    ("#{args} => #{ret}").gsub("Contracts::", "")
+    ("#{args} => #{ret}").gsub("Contracts::Builtin::", "")
```
waterlink: This is good change

lib/contracts/call_with.rb
```diff
@@ -26,7 +26,7 @@ def call_with(this, *args, &blk)
                                                   :return_value => false)
         end
 
-        if contract.is_a?(Contracts::Func)
+        if contract.is_a?(Contracts::Builtin::Func)
```
waterlink: Same here

waterlink: In this file

lib/contracts/formatters.rb
```diff
@@ -90,7 +90,7 @@ def full?
 
       def plain?
         # Not a type of contract that can have a custom to_s defined
-        !@value.is_a?(CallableClass) && @value.class != Class
+        !@value.is_a?(Builtin::CallableClass) && @value.class != Class
```
waterlink: Same here

lib/contracts/formatters.rb
```diff
@@ -103,7 +103,7 @@ def empty_to_s?
       end
 
       def strip_prefix(val)
-        val.gsub(/^Contracts::/, "")
+        val.gsub(/^Contracts::Builtin::/, "")
```
waterlink: This is good

lib/contracts/validators.rb
```diff
@@ -37,13 +37,13 @@ module Validators
         end
       end,
 
-      Contracts::Args => lambda do |contract|
+      Contracts::Builtin::Args => lambda do |contract|
```
waterlink: Same in this file

spec/builtin_contracts_spec.rb
```diff
@@ -356,7 +356,7 @@
     it "can't be used outside of KeywordArgs" do
       expect do
         BareOptionalContractUsed.new.something(3, 5)
-      end.to raise_error(ArgumentError, Contracts::Optional::UNABLE_TO_USE_OUTSIDE_OF_OPT_HASH)
+      end.to raise_error(ArgumentError, Contracts::Builtin::Optional::UNABLE_TO_USE_OUTSIDE_OF_OPT_HASH)
```
waterlink: Same

spec/contracts_spec.rb
```diff
@@ -84,12 +84,12 @@ def no_args_bad_contract
     it "should fail if multiple methods are defined with the same contract (for pattern-matching)" do
       expect do
         Class.new(GenericExample) do
-          Contract Contracts::Num => Contracts::Num
+          Contract Contracts::Builtin::Num => Contracts::Builtin::Num
```
waterlink: same

spec/fixtures/fixtures.rb
```diff
@@ -5,7 +5,7 @@
 class A
   include Contracts::Core
 
-  Contract C::Num => C::Num
+  Contract C::Builtin::Num => C::Builtin::Num
```
waterlink: same

spec/module_spec.rb
```diff
@@ -1,7 +1,7 @@
 module Mod
   include Contracts::Core
 
-  Contract C::Num => C::Num
+  Contract C::Builtin::Num => C::Builtin::Num
```
waterlink: same

spec/override_validators_spec.rb
```diff
@@ -21,7 +21,7 @@
       klass = Class.new do
         include Contracts::Core
 
-        Contract ({ :a => Contracts::Num, :b => String }) => nil
```
waterlink: same

TUTORIAL.md
```diff
@@ -579,8 +579,8 @@ Possible validator overrides:
 - `override_validator(Hash)` - e.g. `{ :a => C::Num, :b => String }`,
 - `override_validator(Range)` - e.g. `(1..10)`,
 - `override_validator(Regexp)` - e.g. `/foo/`,
-- `override_validator(Contracts::Args)` - e.g. `C::Args[C::Num]`,
-- `override_validator(Contracts::Func)` - e.g. `C::Func[C::Num => C::Num]`,
+- `override_validator(Contracts::Builtin::Args)` - e.g. `C::Builtin::Args[::BuiltinC::Num]`,
```
waterlink: This line has a typo in the example.

TUTORIAL.md
waterlink: If I think about `validator` list, it does not need to have `Builtin::` in it too.

waterlink:
@PikachuEXE Looks very good so far. Left some comments around `Builtin::` usage in docs and code. 

Ideally, since `include Builtin` have been added in `Contracts`, we don't need to use `Builtin::` anywhere, nor our users need to know about `Contracts::Builtin::*`, they need to know only about `include Contracts::Core` and `include Contracts::Builtin` and about `Contracts::ContractName` and `ContractName` shortcut (in case `Contracts::Builtin` is included).

lib/contracts/formatters.rb
PikachuEXE: This is required or all spec will fail
(`#<NameError: uninitialized constant Contracts::Formatters::InspectWrapper::CallableClass>`)

PikachuEXE:
Most changes reverted except a few places
See my line note(s)

waterlink:
LGTM

waterlink:
@PikachuEXE Thanks!

PikachuEXE:
@waterlink Do you know when will this & #198 be released?

waterlink:
I will create a PR later today with changelog update and version bump. Then it is up to @egonSchiele to merge & release it to rubygems.

BTW, there is still need to add info on `include Contracts::Builtin` to `TUTORIAL.md` in `Built-in contracts` section.

waterlink:
I will add this info later, no need to bother :)

waterlink:
created

PR #210, opened 2015-11-18
Contracts on singleton classes of subclasses do not work

betamatt:
Bug report in the form of a failing test.

Given:
```
class A
  include Contracts::Core
end

class B < A
  class << self
    Contract Integer => Integer
    def num(i); i; end
  end
end
```

You're going to get `Contracts::ContractsNotIncluded: In order to use contracts in singleton class, please include Contracts module in original class`

### DIFF
--- a/spec/contracts_spec.rb
+++ b/spec/contracts_spec.rb
@@ -176,6 +176,23 @@ RSpec.describe "Contracts:" do
     end
   end
 
+  describe "usage in the singleton class of a subclass" do
+    subject {
+      Class.new(GenericExample) do
+        class << self
+          Contract Integer => Integer
+          def num(int)
+            int
+          end
+        end
+      end
+    }
+
+    it "should work with a valid contract on a singleton method" do
+      expect(subject.num(1)).to eq(1)
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


egonSchiele:
Thanks for the PR @betamatt ! Could you fix ci?

betamatt:
@egonSchiele ha, well, funny story. This isn't a real PR, it's just illustrating a problem that I haven't had a chance to dig into fixing. I figured a test case was more useful than an empty bug report.

egonSchiele:
Aha, I didn't check the PR. Thanks for the bug report!

waterlink:
@egonSchiele @betamatt I am taking closer look into this.

waterlink:
Heh, We actually have test that verifies that exactly in that case error happens: https://github.com/betamatt/contracts.ruby/blob/feature/subclass-singleton/spec/contracts_spec.rb#L144-L168 :)

waterlink:
Since I was able to fix that case I think I might as well remove that test.

waterlink:
I will create new PR and will hope test pass for 1.8.7 :)

PR #211, opened 2015-11-23
Allow contracts on singleton class of subclass

waterlink:
fixes #210

### DIFF
--- a/lib/contracts/decorators.rb
+++ b/lib/contracts/decorators.rb
@@ -4,6 +4,10 @@ module Contracts
       Engine.apply(klass)
     end
 
+    def inherited(subclass)
+      Engine.fetch_from(subclass).set_eigenclass_owner
+    end
+
     def method_added(name)
       MethodHandler.new(name, false, self).handle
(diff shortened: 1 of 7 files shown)


waterlink:
Oh that is so meta. I have run into problems with our feature (cucumber) specs for `contracts.ruby` because I have broken functionality that `cucumber/aruba` uses from `contracts.ruby`.

waterlink:
Finally made this pass on all rubies.

@betamatt @egonSchiele Can you take a look?

egonSchiele:
Looks great to me, thanks @waterlink ! Lets see if @betamatt has any comments, otherwise I will merge this.

betamatt:
Thanks a lot.

I haven't studied enough on how you've got this wired so I can't judge the implementation. However, I'll see if I can find some time today to try it out on our codebase and verify that there aren't any obvious regressions.

egonSchiele:
In that case I have just merged it. Thanks @betamatt and @waterlink !

PR #212, opened 2015-11-26
Number fields

### LINKED ISSUE #204: Number groups builtin contracts.
Unfortunately there is no mathematical consensus on what a Natural number is. It's either includes zero or does not depending who you ask. In one notation a + is used to disambiguate (my lecturer did this):

N = N^0 = N_0 = { 0, 1, 2, 3, ... }
N* = N+ = N_1 = N>0 = { 1, 2, 3, ... }

This is basically what we are doing with the `Nat` builtin contract. I would like to add a contract that doesn't include zero for completeness, so if we stick to the current definition of Nat then the contract to be added might be `NatP` for "Natural Positive".

There is an alternate notation I have seen that flips it around and uses W for the zero inclusive set for "Whole Numbers":

W = { 0, 1, 2, 3, ... }
N = { 1, 2, 3, ... }

This notation seems nonstandard though fairly widespread still. It would mean we'd have to change the definition of `Nat` then add `Whole` that replaced the old Nat.

We're also missing an `Int` contract. If we have `Num` as a shorthand for Numeric then we should have `Int` as a shorthand for `Integer`.

What do you think? I could open a pull request once we've decided.

sfcgeorge:
Fixes #204 by adding builtin `Int` and `NatPos` contracts.

Added to TUTORIAL.md and fixed documentation links by namespacing builtin.

Added to cucumber features and tested still passing.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -68,44 +68,44 @@ This can be useful if you're in a REPL and want to figure out how a function sho
 contracts.ruby comes with a lot of built-in contracts, including the following:
 
 * Basic types
-  * [`Num`](http://www.rubydoc.info/gems/contracts/Contracts/Num) – checks that the argument is `Numeric`
-  * [`Pos`](http://www.rubydoc.info/gems/contracts/Contracts/Pos) – checks that the argument is a positive number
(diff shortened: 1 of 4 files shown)


waterlink:
LGTM

egonSchiele:
THanks @sfcgeorge !

PR #216, opened 2016-01-09
Add ruby 2.3 to travis

waterlink:


waterlink:
Ok, it is green, will squash it.

waterlink:
The trick of installing ruby 2.3 on travis is to upgrade to `trusty` beta build. The downside is that ruby 1.8, 1.9.2 and jrubies can't be used yet for `trusty` build. Build matrix comes to the rescue in this PR.

@egonSchiele Can you take a look?

egonSchiele:
lgtm @waterlink !

.travis.yml
```diff
@@ -1,13 +1,12 @@
+dist: trusty
```
PikachuEXE: Does it matter?
What is the default value? (with reference)

PikachuEXE:
Oops forgot to read comments first :P
But I can [put `2.3.0` without using `trusty`](https://github.com/fnando/i18n-js/commit/d0e698e4dca2579b02b2855adab077f18dcb03b8)
Also surprisingly using alias does not always point to the latest patch level MRI
An example would be [using `2.2` points to `2.2.0` not `2.2.4`](https://travis-ci.org/fnando/i18n-js/jobs/100535252)

waterlink:
That happens because of `--fuzzy` option, that travis adds to the `bundler`
command

--
Best Regards,
Oleksii Fedorov,
+49 15757 486 476

PR #224, opened 2016-02-02
Add syntax highlighting for error message

adarsh:
Reason for change:
* The `README.md` was reading a bit weird without the syntax highlighting.

Change:
* Wrap the example error message in triple backticks.


waterlink:
@adarsh Thanks for noticing! It was a twice-indented kind of code-block, but somehow after some edit, first line was dedented (probably by me :) ), effectively disabling code-block formatting.

LGTM

PR #227, opened 2016-02-23
Adds DescendantOf

naps62:
This adds the ability to make assertions on the class hierarchy of a class being passed as an argument. It should be useful when using things like the Adapter Pattern or Dependency Injection.

Consider the following example:

```ruby
class SlackClientAdapter < BaseClient
end

class HipchatClientAdapter < BaseClient
end

Contract C::DescendantOf[BaseClient] => BaseClient
def start_client(client_class)
  @client = client_class.new
end

start_client(SlackClientAdapter)
start_client(HipchatClientAdapter)
```

This is a demonstration of what I'm actually doing on a real app (although the example itself is totally made up).

I want to be able to assert that the argument is not just a class, but a descendant of `BaseClient` (this base client serves as an abstract class this example. God, I feel like a Java developer all over again...)

I'm not sure if this is something that's already been proposed and discarded, so I didn't take a lot of care in the organization of the code I added.
Should this be accepted, I'll gladly make any required changes before merging.

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -420,6 +420,30 @@ module Contracts
       attr_reader :options
     end
 
+    # Use this for specifying contracts for class arguments
+    # Example: <tt>Descendant[ e: Range, f: Optional[Num] ]</tt>
+    class DescendantOf < CallableClass
+      def initialize(parent_class)
+        @parent_class = parent_class
+      end
+
+      def valid?(given_class)
+        given_class.is_a?(Class) && given_class.ancestors.include?(parent_class)
+      end
+
+      def to_s
+        "DescendantOf[#{parent_class}]"
+      end
+
+      def inspect
+        to_s
+      end
+
+      private
+
+      attr_reader :parent_class
+    end
+
     # Use this for specifying optional keyword argument
     # Example: <tt>Optional[Num]</tt>
     class Optional < CallableClass
--- a/spec/builtin_contracts_spec.rb
+++ b/spec/builtin_contracts_spec.rb
@@ -3,6 +3,24 @@ RSpec.describe "Contracts:" do
     @o = GenericExample.new
   end
 
+  describe "DescendantOf:" do
+    it "should pass for Array" do
+      expect { @o.enumerable_descendant_test(Array) }.to_not raise_error
+    end
+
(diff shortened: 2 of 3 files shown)


egonSchiele:
Hi @naps62 ,
Actually, that's how things currently work...you don't need `DescendantOf` for this functionality. For example:

```ruby
require "contracts"
include Contracts

class Animal; end
class Dog < Animal; end
class Cat < Animal; end

Contract Animal => Any
def foo x
  p x
end

foo Dog.new # works, Dog is an Animal (checked with is_a?)
foo "hello" # fails, String is not an Animal
```

The check happens [here](https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts/validators.rb#L57).

If you want to do the opposite (i.e. accept `Animal` only, not `Dog` or `Cat`), use `Exactly`:

```ruby
Contract Exactly[Animal] => Any
def foo x
  p x
end
```

waterlink:
Not exactly, here the argument is not an instance of class, but class itself. 

Dog.new as opposed to just Dog.

naps62:
@waterlink is right. following @egonSchiele's example, the use case for this would be something like:

```
Contract DescendantOf[Animal] => Animal
def animal_factory(x)
  x.new
end

dog = animal_factory(Dog)
```

egonSchiele:
Oops, I misread! You're right, that is useful. Could you fix ci, then I'll merge it:

```
lib/contracts/builtin_contracts.rb:431:9: C: Style/RedundantReturn: Redundant return detected.
        return given_class.ancestors.include?(@parent_class)
```

naps62:
@egonSchiele ok great. The build is now fixed

egonSchiele:
Thank you for the PRs!

lib/contracts/builtin_contracts.rb
```diff
@@ -420,6 +420,30 @@ def inspect
       attr_reader :options
     end
 
+    # Use this for specifying contracts for class arguments
+    # Example: <tt>Descendant[ e: Range, f: Optional[Num] ]</tt>
```
PikachuEXE: Should this line be `DescendantOf` instead?

egonSchiele: You are correct @PikachuEXE. Fix is on master.

Issue #209, opened 2015-10-15
KeywordArgs causes error when used with Func


rap-kasta:
Code to replicate problem:

```ruby
require 'contracts'
include Contracts::Core
include Contracts::Builtin

Contract KeywordArgs[name: Optional[String]], Func[String => String] => String
def hey_you(name: 'John', &block)
  "Hey, #{yield name}!"
end

puts hey_you{|name| name.reverse }
puts hey_you(name: 'George'){|name| name.reverse }
```

Raised error:

```
/Users/rk/.rvm/gems/ruby-2.2.3@soft24/gems/contracts-0.12.0/lib/contracts/builtin_contracts.rb:390:in `valid?': undefined method `keys' for #<Proc:0x007fbf8a16ba60@keyword_args_vs_func.rb:10> (NoMethodError)
	from /Users/rk/.rvm/gems/ruby-2.2.3@soft24/gems/contracts-0.12.0/lib/contracts/validators.rb:53:in `block (2 levels) in <module:Validators>'
	from /Users/rk/.rvm/gems/ruby-2.2.3@soft24/gems/contracts-0.12.0/lib/contracts/call_with.rb:18:in `[]'
	from /Users/rk/.rvm/gems/ruby-2.2.3@soft24/gems/contracts-0.12.0/lib/contracts/call_with.rb:18:in `block in call_with'
	from /Users/rk/.rvm/gems/ruby-2.2.3@soft24/gems/contracts-0.12.0/lib/contracts/call_with.rb:13:in `times'
	from /Users/rk/.rvm/gems/ruby-2.2.3@soft24/gems/contracts-0.12.0/lib/contracts/call_with.rb:13:in `call_with'
	from /Users/rk/.rvm/gems/ruby-2.2.3@soft24/gems/contracts-0.12.0/lib/contracts/method_handler.rb:138:in `block in redefine_method'
	from keyword_args_vs_func.rb:10:in `<main>'
```

egonSchiele:
I can reproduce. `puts hey_you{|name| name.reverse }` raises the error. Optional args + blocks seem to lead to many issues.

egonSchiele:
PR is here https://github.com/egonSchiele/contracts.ruby/pull/230

egonSchiele:
Fixed

PR #230, opened 2016-03-07
Fix for issue #209

egonSchiele:
The issue here was, the arguments to `KeywordArgs` are optional. So if they are not provided, the params and contracts don't line up correctly.

```
KeywordArgs    Proc
{}                     <Proc>
```

becomes

```
KeywordArgs    Proc
<Proc>
```

Fix is in this commit. Could you review @waterlink ?
cc @rap-kasta

### DIFF
--- a/spec/contracts_spec.rb
+++ b/spec/contracts_spec.rb
@@ -521,16 +521,6 @@ RSpec.describe "Contracts:" do
     end
   end
 
-  describe "keyword args with defaults, with a block" do
-    it "should work when both keyword args and a block is given" do
-      expect(@o.keyword_args_hello(name: "maggie", &:upcase)).to eq("Hey, MAGGIE!")
-    end
-
-    it "should work even when keyword args aren't given" do
-      expect(@o.keyword_args_hello(&:upcase)).to eq("Hey, ADIT!")
-    end
-  end
-
   describe "classes" do
     it "should not fail for an object that is the exact type as the contract" do
       p = Parent.new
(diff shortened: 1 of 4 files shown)


waterlink:
Looks good to me. Is it possible to add a failing spec for this?

egonSchiele:
Good call -- done. Will squash and merge.

spec/contracts_spec.rb
```diff
@@ -521,6 +521,16 @@ def self.greeting(name)
...
+  describe "keyword args with defaults, with a block" do
+    it "should work when both keyword args and a block is given" do
+      expect(@o.keyword_args_hello(name: "maggie", &:upcase)).to eq("Hey, MAGGIE!")
+    end
+
+    it "should work even when keyword args aren't given" do
+      expect(@o.keyword_args_hello(&:upcase)).to eq("Hey, ADIT!")
+    end
+  end
+
```
waterlink: These tests need to go to ruby-specific tests.

spec/fixtures/fixtures.rb
```diff
@@ -123,6 +123,11 @@ def person_keywordargs(data)
   def hash_keywordargs(data)
   end
 
+  Contract C::KeywordArgs[name: C::Optional[String]], C::Func[String => String] => String
+  def keyword_args_hello(name: "Adit", &block)
+    "Hey, #{yield name}!"
+  end
+
```
waterlink: Same here, this definition needs to go to ruby-specific spec file.

Issue #215, opened 2016-01-09
Include input parameters in ReturnContractError


beezee:
Because these are runtime, I think it's way more useful to see what input produced an invalid return, considering it doesn't necessarily mean you have this error on all possible inputs.

Is this something folks agree with, or are there concerns about it?

If it would be accepted (barring any code review of course) I would be happy to open a PR

egonSchiele:
Sounds good to me, send over a PR!

alex-fedorov:
:+1: Sounds good to me too!

--
Best Regards,
Oleksii Fedorov,
Sr Ruby, Clojure, Crystal, Golang Developer,
Microservices Backend Engineer,
+49 15757 486 476

PR #231, opened 2016-03-08
ContractError should also show the arguments that caused the error

egonSchiele:
Output:

```
contracts.ruby (adit/215) $ cat test.rb
require "contracts"
include Contracts

Contract Num => Num
def incr x
  "Sdf"
end

incr(23)
contracts.ruby (adit/215) $ ruby test.rb
/Users/abhargava/contracts.ruby/lib/contracts.rb:45:in `block in <class:Contract>': Contract violation for return value: (ReturnContractError)
        Expected: Num,
        Actual: "Sdf"
        Value guarded in: Object::incr
        With Contract: Num => Num
        All arguments: [23]
        At: test.rb:5
	from /Users/abhargava/contracts.ruby/lib/contracts.rb:155:in `call'
	from /Users/abhargava/contracts.ruby/lib/contracts.rb:155:in `failure_callback'
	from /Users/abhargava/contracts.ruby/lib/contracts/call_with.rb:82:in `call_with'
	from /Users/abhargava/contracts.ruby/lib/contracts/method_handler.rb:138:in `block in redefine_method'
	from test.rb:9:in `<main>'
```

cc @waterlink @beezee (for issue #215)

### DIFF
--- a/lib/contracts.rb
+++ b/lib/contracts.rb
@@ -111,7 +111,7 @@ class Contract < Contracts::Decorator
     ("#{args} => #{ret}").gsub("Contracts::Builtin::", "")
   end
 
-  # Given a hash, prints out a failure message.
+  # Given a hash, returns a failure message.
   # This function is used by the default #failure_callback method
   # and uses the hash passed into the failure_callback method.
   def self.failure_msg(data)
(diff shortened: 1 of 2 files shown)


egonSchiele:
Oh man, I forgot I need to update the output in the Cucumber tests :/

waterlink:
This is a very useful feature

Issue #229, opened 2016-02-24
Weird errors with included modules


naps62:
I'm trying to add Contracts to class methods defined in a module. This is what I currently have working:

```ruby
class MyClass
  include MyModule

  include Contracts::Core
  include Contracts::Builtin

  # some other methods here, with contracts
end

module MyModule
  def self.included(base)
    base.extend ClassMethods
  end

  module ClassMethods
    def a_method
    end
end
```

This works (I can call `MyClass.a_method`)
But I can't seem to add contracts to the module without messing up the class where it is included. For example:

```ruby
module MyModule
  def self.included(base)
    base.extend ClassMethods
  end

  module ClassMethods
    include ::Contracts::Core

    def a_method
    end
end
```

the mere inclusion of Contracts::Core on ClassMethods messes up the contracts on `MyClass`, which now doesn't compile because `'Contract': undefined method 'Contract' for Class:Class (NoMethodError)`

Is there any solution to this that doesn't involve me having to use the `C = Contracts` shortcut everywhere instead of including the modules?

naps62:
Update: it turns out the `C = Contracts` shorcut doesn't work either, because it only provides access to the Builtin contracts, no the `Contract` method itself

So I'm officially unable to have contracts in the ClassMethods module without causing NoMethodError's on the class that is extended by it

waterlink:
Hi @naps62! Thanks for the bug report.

Could you please make a small gist (on https://gist.github.com), that is a runnable code and re-produces the bug? Either just one file, or one file + spec file.

naps62:
@waterlink Here you go: https://gist.github.com/naps62/7ee35e53635e1b7ee9ae

Since I first posted the issue, I found that I can make it work by simply changing the order in which I include modules (this is what's reproduced in the gist). It's still a bug I guess, although not as serious

egonSchiele:
Hey @naps62 and @waterlink ,
There's an annoying reason for why this happens. Because the `Contract` function starts with an uppercase letter, Ruby treats is as a constant. And it treats constants differently. Here's a simple version of what's happening:

```ruby
class Foo
  def self.lowercase
    puts "lowercase"
  end

  def self.Uppercase
    puts "uppercase method"
  end
end

Foo::lowercase
Foo::Uppercase
```

Running this code prints out:

```
lowercase
foo.rb:12:in `<main>': uninitialized constant Foo::Uppercase (NameError)
```

[...]

PR #232, opened 2016-03-08
issue #229: fix constant lookup when including a module that already includes Contracts

egonSchiele:
Based on my comment in issue #229


waterlink:
LGTM. Feel free to merge when green :)

PR #234, opened 2016-03-17
Fixed issue where KeywordArgs is not checked if it is a hash.

dpad:
This leads to issues when methods are overloaded, because when the
KeywordArgs gets checked on a non-hash argument, it raises an exception
rather than failing properly.

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -401,6 +401,7 @@ module Contracts
       end
 
       def valid?(hash)
+        return false unless hash.is_a?(Hash)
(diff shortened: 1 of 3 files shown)


waterlink:
Thanks @dpad

Could you please add this edge case to our test suite?

waterlink:
For example in this spec file https://github.com/egonSchiele/contracts.ruby/blob/master/spec/ruby_version_specific/contracts_spec_2.0.rb

egonSchiele:
Tests would be great. Thank you for the fix @dpad !

dpad:
Added a test that fails without the fix, succeeds with.

Issue #235, opened 2016-04-13
Strict Hash?


cbrwizard:
Hey, I've just started using this gem and I have noticed that when writing a contract for a hash 'You don't need to put a contract on every key'.
That is the contrary to what I need. I would like to define a strict schema of hash data and if a hash with additional keys is used, there should be a failure.
Is there a way to do that easily with this gem? Thanks!

waterlink:
Sounds like you want to have a custom contract. Here is how you create one:

```ruby
class ExactlyHash
  def initialize(hash)
    @hash = hash
  end

  def valid?(arg)
    hash.all? do |key, contract|
      arg.has_key?(key) && Contract.valid?(arg[key], contract)
  end
end
```

And use it as `Contract ExactlyHash.new({ ... })`

waterlink:
Sorry if code might not work, I am from the phone.

cbrwizard:
@waterlink thank you for the code! It didn't work but I was able to quickly fix it based on your work.
Want me to add a pull request and add this Contract to the repo? Someone else might need this as well.

egonSchiele:
Please do @cbrwizard ! Add it to this file:
https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts/builtin_contracts.rb

Have your class inherit `CallableClass`:
https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts/builtin_contracts.rb#L87

That way you can use it as `StrictHash[...]` instead of `StrictHash.new(...)`.

cbrwizard:
Hey @egonSchiele I do not mean to push you or anything, but I'm just curious: is there a time estimate when StrictHash is going to be in the gem? I might need it in production soon =] I have tried to tell bundler to take code from github but for some reason it didn't download the last version. I am more a JS person than Ruby so sorry if I made a mistake here.

egonSchiele:
@cbrwizard I just pushed it to rubygems! Look for v0.14.0.

cbrwizard:
@egonSchiele awesome! It works in my project. I guess we can close this issue now

PR #236, opened 2016-04-18
StrictHash

cbrwizard:
Adds a StrictHash builtin contract, which checks that the argument is
a hash, and every key passed is present in the given contract.

https://github.com/egonSchiele/contracts.ruby/issues/235

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -393,6 +393,25 @@ module Contracts
       end
     end
 
+    # Use this to specify the Hash characteristics. This contracts fails
+    # if there are any extra keys that don't have contracts on them.
+    # Example: <tt>StrictHash[{ a: String, b: Bool }]</tt>
+    class StrictHash < CallableClass
+      attr_reader :contract_hash
+
+      def initialize(contract_hash)
+        @contract_hash = contract_hash
+      end
+
+      def valid?(arg)
+        return false unless arg.is_a?(Hash)
+
+        contract_hash.all? do |key, _v|
+          contract_hash.key?(key) && Contract.valid?(arg[key], contract_hash[key])
+        end
+      end
+    end
+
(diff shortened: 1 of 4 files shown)


cbrwizard:
If there are any issues - please tell me, I will be glad to fix/adjust them!

lib/contracts/builtin_contracts.rb
```diff
@@ -393,6 +393,25 @@ def validate_hash(hash)
       end
     end
 
+    # Use this to specify the Hash characteristics with strict check of keys'
```
egonSchiele: Could you elaborate on "strict check of keys' presence"? It would be nice to say "the contracts fails if there are any extra keys that don't have contracts on them".

spec/builtin_contracts_spec.rb
```diff
@@ -418,4 +418,41 @@ def something(hash)
       end
     end
   end
+
+  describe "StrictHash:" do
+    context "when given an exact correct input" do
+      it "does not raise an error" do
+        expect { @o.strict_person(:name => "calvin", :age => 10) }
```
egonSchiele: These lines are causing a syntax error in CI with Ruby 1.8.7: 
https://travis-ci.org/egonSchiele/contracts.ruby/jobs/123853792

You could do this as a block instead.

egonSchiele:
@cbrwizard this looks awesome! Thank you for adding documentation and tests. I left two small comments for you, it is good to merge after those are fixed.

spec/builtin_contracts_spec.rb
cbrwizard: I couldn't even install 1.8.7 properly to test it locally heh...

cbrwizard:
@egonSchiele thanks, here is an update!

egonSchiele:
Merged, thanks!

PR #237, opened 2016-05-17
Fix contracts used in AR-models

### LINKED ISSUE #223: ActiveRecord (3.2.22) shows deprecation warning due to override of .inherited method
Given some (even empty) model:

```ruby
class Book < ActiveRecord::Base
  include Contracts::Core
end
```

`ActiveRecord` puts warning (not so promising for actual versions of AR…):

> DEPRECATION WARNING: It looks like something (probably a gem/plugin) is overriding the ActiveRecord::Base.inherited method. It is important that this hook executes so that your models are set up correctly. A workaround has been added to stop this causing an error in 3.2, but future versions will simply not work if the hook is overridden. If you are using Kaminari, please upgrade as it is known to have had this problem.
> 
> The following may help track down the problem: ["/home/rk/.rvm/gems/ruby-2.2.2@soft24/gems/contracts-0.13.0/lib/contracts/decorators.rb", 7]

In `contracts/decorators.rb:7` method definition there are actually no call to `super`.

eval:
ActiveRecord::Base.inherited needs to be called in order for models to be setup correctly.

Fixes #223

### DIFF
--- a/lib/contracts/decorators.rb
+++ b/lib/contracts/decorators.rb
@@ -6,6 +6,7 @@ module Contracts
 
     def inherited(subclass)
       Engine.fetch_from(subclass).set_eigenclass_owner
+      super
     end
 
     def method_added(name)


waterlink:
Thanks @eval

PR #245, opened 2016-12-20
make specs less noisy to read by using declarative expectations

mindreframer:
problem: 
- specs are a bit hard to read

solution: 
- use `fails` / `passes` methods to indicate desired expectation 
- specs become more declarative and descriptive
- the changes are of mechanical nature, no new logic is introduced

### DIFF
--- a/spec/builtin_contracts_spec.rb
+++ b/spec/builtin_contracts_spec.rb
@@ -3,296 +3,303 @@ RSpec.describe "Contracts:" do
     @o = GenericExample.new
   end
 
+  def fails(&some)
+    expect { some.call }.to raise_error(ContractError)
+  end
+  def passes(&some)
+    expect { some.call }.to_not raise_error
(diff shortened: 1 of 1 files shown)


spec/builtin_contracts_spec.rb
```diff
@@ -3,296 +3,303 @@
     @o = GenericExample.new
   end
 
+  def fails(&some)
+    expect { some.call }.to raise_error(ContractError)
+  end
+  def passes(&some)
```
egonSchiele: Can you add a blank line before the `def passes(&some)` ? Currently this PR is failing because rubocop is complaining about the missing blank line: https://travis-ci.org/egonSchiele/contracts.ruby/jobs/185597019

egonSchiele:
Nice, lgtm. You need to add a blank line, I will merge this after that.

PR #246, opened 2016-12-21
custom error formatter for KeywordArgs

mindreframer:
problem: 
- it is impossible to quickly understand the contract violation with big hashes 

solution: 
- provide custom error formatter just for keywordargs, that gives you extra information about: 
- Missing Contract definition (for unspecified keys in contract)
- Invalid Args (validation failures)
- Missing Args (required values, that are missing)

### DIFF
--- a/lib/contracts/error_formatter.rb
+++ b/lib/contracts/error_formatter.rb
@@ -1,0 +1,124 @@
+module Contracts
+  class ErrorFormatters
+    def self.failure_msg(data)
+      class_for(data).new(data).message
+    end
+
+    def self.class_for(data)
+      return Contracts::KeywordArgsErrorFormatter if is_keysword_args?(data)
+      return DefaultErrorFormatter
+    end
+
+    def self.is_keysword_args?(data)
+      data[:contract].is_a?(Contracts::Builtin::KeywordArgs) && data[:arg].is_a?(Hash)
+    end
+  end
+
+  class DefaultErrorFormatter
+    include Contracts::Colorize
+    attr_accessor :data
+    def initialize(data)
+      @data = data
+    end
+
+    def message
+      %{#{header}
+        Expected: #{expected},
+        Actual: #{data[:arg].inspect}
+        Value guarded in: #{data[:class]}::#{method_name}
+        With Contract: #{data[:contracts]}
+        At: #{green(position)} }
+    end
+
+  private
+    def header
+      if data[:return_value]
+         "Contract violation for return value:"
+       else
+         "Contract violation for argument #{data[:arg_pos]} of #{data[:total_args]}:"
+       end
+    end
+
+    def expected
+      Contracts::Formatters::Expected.new(data[:contract]).contract
+    end
+
+    def position
+      Contracts::Support.method_position(data[:method])
+    end
+
+    def method_name
+      Contracts::Support.method_name(data[:method])
+    end
+  end
+
+  class KeywordArgsErrorFormatter < DefaultErrorFormatter
+    def message
+      s = []
+      s << header
+      s << "Expected: #{expected}"
+      s << "Actual: #{data[:arg].inspect}"
+      s << "Missing Contract: #{missing_contract_info}" if !missing_contract_info.empty?
+      s << "Invalid Args: #{invalid_args_info}"         if !invalid_args_info.empty?
+      s << "Missing Args: #{missing_args_info}"         if !missing_args_info.empty?
+      s << "Value guarded in: #{data[:class]}::#{method_name}"
+      s << "With Contract: #{data[:contracts]}"
+      s << "At: #{position}"
+
+      s.join("\n")
+    end
+
+    private
+
+    def missing_args_info
+      @missing_args_info ||= begin
+        missing_keys = contract_options.keys - arg.keys
+        contract_options.select do |key, value|
+          missing_keys.include?(key)
+        end
+      end
+    end
+
+    def missing_contract_info
+      @missing_contract_info ||= begin
+        contract_keys = contract_options.keys
+        arg.select{|key, value| !contract_keys.include?(key)}
+      end
+    end
+
+    def invalid_args_info
+      @invalid_args_info ||= begin
+        invalid_keys = []
+        arg.each do |key, value|
+          if contract = contract_options[key]
+            if !(check_contract(contract, value))
+              invalid_keys.push(key)
+            end
+          end
+        end
+        invalid_keys.map do |key|
+          {key => arg[key], :contract => contract_options[key] }
+        end
+      end
+    end
+
+    def check_contract(contract, value)
+      if contract.respond_to?(:valid?)
+        contract.valid?(value)
+      else
+        value.is_a?(contract)
+      end
+    rescue
+      false
+    end
+
+
+    def contract_options
+      @contract_options ||= data[:contract].send(:options)
+    end
+
+    def arg
+      data[:arg]
+    end
+  end
+end
--- a/spec/error_formatter_spec.rb
+++ b/spec/error_formatter_spec.rb
@@ -1,0 +1,64 @@
+RSpec.describe "Contracts::ErrorFormatters" do
+  before :all do
+    @o = GenericExample.new
+  end
+
+  describe "self.class_for" do
+    it "returns the right formatter for passed in data" do
+    end
+  end
+
+  def format_message(str)
+    str.split("\n").map(&:strip).join("\n")
+  end
+
+  describe "self.failure_msg" do
+    it "includes normal information" do
+      expect do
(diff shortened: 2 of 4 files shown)


rous-gg:
This is a really good feature to have. Right now it's really hard to understand what is wrong with keyword args contract and you spend a lot of time making that diff in mind.

mindreframer:
here is an example how the error message would look like: 


```
ParamContractError:
       Contract violation for argument 1 of 1:
       Expected: (KeywordArgs[{:guid=>String, :occurred_at=>DateTime, :created_by_id=>Contracts::Builtin::Num}])
       Actual: {:guid=>"4fea1b56-3218-4c4c-a2ce-6db42d8e445f", :occurred_at=>"1", :test1=>1}
       Missing Contract: {:test1=>1}
       Invalid Args: [{:occurred_at=>"1", :contract=>DateTime}]
       Missing Args: {:created_by_id=>Contracts::Builtin::Num}
       Value guarded in: Events::AbstractEvent::initialize
       With Contract: KeywordArgs => NilClass

```

lib/contracts/error_formatter.rb
```diff
@@ -0,0 +1,121 @@
...
+    def self.failure_msg(data)
+      class_for(data).new(data).message
+    end
+
+    def self.class_for(data)
+      return Contracts::KeywordArgsErrorFormatter if keysword_args?(data)
+      DefaultErrorFormatter
+    end
+
+    def self.keysword_args?(data)
```
egonSchiele: typo, should be `keyword_args`

lib/contracts/error_formatter.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      s.join("\n")
+    end
+
+    private
+
+    def missing_args_info
+      @missing_args_info ||= begin
+        missing_keys = contract_options.keys - arg.keys
+        contract_options.select do |key, _value|
```
egonSchiele: you can use `_` instead of `_value` since that var isn't used

lib/contracts/error_formatter.rb
```diff
@@ -0,0 +1,121 @@
...
+          missing_keys.include?(key)
+        end
+      end
+    end
+
+    def missing_contract_info
+      @missing_contract_info ||= begin
+        contract_keys = contract_options.keys
+        arg.select { |key, _value| !contract_keys.include?(key) }
```
egonSchiele: same

lib/contracts/error_formatter.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+    end
+
+    def check_contract(contract, value)
+      if contract.respond_to?(:valid?)
+        contract.valid?(value)
+      else
+        value.is_a?(contract)
+      end
+    rescue
```
egonSchiele: why was this rescue needed?

lib/contracts/error_formatter.rb
```diff
@@ -0,0 +1,121 @@
...
+        end
+        invalid_keys.map do |key|
+          {key => arg[key], :contract => contract_options[key] }
+        end
+      end
+    end
+
+    def check_contract(contract, value)
+      if contract.respond_to?(:valid?)
+        contract.valid?(value)
```
egonSchiele: I don't like the idea of calling the `valid` method of a contract again, to display the error message, for a couple of reasons:

1. It breaks the idea that contracts are only run once...so if someone happens to write a contract that takes a long time to check, they wouldn't expect the error message to take time to render.

2. Error rendering should ideally be very simple. It shouldn't be possible to generate an error *while* rendering the error message. But code like this means we'll be running arbitrary code by other folks (anyone can write their own class with a valid? functions), which could fail for whatever reason. Maybe that is why you added the `rescue`?

spec/error_formatter_spec.rb
```diff
@@ -0,0 +1,57 @@
+RSpec.describe "Contracts::ErrorFormatters" do
+  before :all do
+    @o = GenericExample.new
+  end
+
+  describe "self.class_for" do
+    it "returns the right formatter for passed in data" do
```
egonSchiele: this block needs a test inside it

lib/contracts/error_formatter.rb
rous-gg: 1. Right now data object does not contain information about specific invalid keyword args keys. it's really hard to make a diff in mind to find them. in most situations contracts are really simple one (type checks, range checks etc) and they do not take much time to execute. Actually we can move that to config so developers will be able to enable/disable this failure msg calculation globally. The main purpose for it is to save time when working in test and dev modes. What do you think if we move that to config?

2. We create a lot of custom contracts in our applications utilising :valid? method. You can both return false in it or raise ParamContractError. This is why we've added rescue to it.

rous-gg: Let's take a look at the example:
       Invalid Args: [{:occurred_at=>"1", :contract=>DateTime}]
One line of output tells us
  - which argument is wrong
  - argument value
  - contract
This should help developers a lot

lib/contracts/error_formatter.rb
mindreframer: done

spec/error_formatter_spec.rb
mindreframer: done

lib/contracts/error_formatter.rb
mindreframer: done

lib/contracts/error_formatter.rb
mindreframer: if you have expensive logic in your contract validators, you are doing it __wrong__. Elixir for example allows only some selected bits of logic to be run in function guards, to make sure that your function call is not burdened by accidental side-effects from guards. Sure, here it is theoretically possible to do anything, but that responsibility should be up to the user of the library. On the other hand, it would be possible to modify the `data` hash to include all the information about failed/missing validations, then we would not need to run validations twice for error message generation... Life is about tradeoffs... @egonSchiele  Which alternative would you prefer?

egonSchiele:
Hey all,
Sorry for my late response here. @mindreframer could you update the data hash? I like that better than running the contracts twice.

mindreframer:
Hey @egonSchiele, good to hear from you! 

Because the discussion here ended kinda abruptly and I liked the idea of contracts for Ruby, I have continued my work on a fork here: https://github.com/ddd-ruby/contracts.ruby

Many changes I have introduced would be too disruptive and would be probably rejected here after a longer discussion, and honestly, I would not even blame you for that :). You as maintainer of this project should be more conservative towards changes. 

I tried to simplify contracts from the DX perspective, that means
- less backwards-compatibility (drop support for rubies 1.8.7- 1.9.3)
- faster tests: remove cucumber tests, because they are kinda fragile and expensive to maintain
- increase the code climate index (now at 3.9), that required quite some refactoring
- reduce cyclomatic complexity as much as possible, that means all the nested if-else blocks were refactored to be much flatter

Overall the resulting codebase should be simpler and a bit easier to maintain / contribute to. One gotcha: the refactoring made it a bit slower, it is roughly 10% slower in trivial benchmarks that the original implementation. 

My current fork satisfies my needs and I have already invested enough efforts into it. If you are interested in some of those changes, you might consider merging some parts yourself...

I really hope this does not appear offensive or rude to you. I just did not have the patience to wait for an undefinite time period for a decision from your side. My brain just does not work like this. :P 

Overall thanks for `Contracts` and maybe you will like some parts of my changes and use them. 

Best, 
Roman

egonSchiele:
No problem Roman, it is not offensive at all :) I don't work in Ruby these days, so it is a bit of overhead for me to maintain the project, which is why my response took a while. I'm really glad to see you like the project ... you have made a ton of commits!

I wish I had more time to work on contracts, I would definitely look at your changes. The codebase is quite complex so I would like to simplify it also.

mindreframer:
@egonSchiele thx for staying cool! If you don't work in Ruby anymore, maybe you could look around for another maintainer / co-maintainer with rights to merge / push to Rubygems? That way contracts would not suffer a slow death... 

A separate Github team would be a nice touch, like:  `https://github.com/contracts.ruby/contracts`
Then you could post a "looking for maintainer(s) for contracts" post on http://www.rubyflow.com/ and link to a github issue with the same title... Some people with more time than you and active Ruby investment might be interested. Pick somebody you would trust to keep the project alive. 

Also there are 23 watchers on the repo. ))) 

But well... That is maybe something you should figure out for yourself :))) 

Best,
Roman

mindreframer:
here is another project where the original maintainer looked for new maintainers: 

https://github.com/fgrehm/vagrant-cachier/issues/143#issuecomment-273523976

egonSchiele:
good idea!

Issue #242, opened 2016-11-23
Func contract's return value isn't enforced with blocks


indigoviolet:
```
Contract Func[Num=>Num] => nil
def foo(&blk)
  result = blk.call(2)
  nil
end

Contract Func[Num=>Num] => nil
def bar(blk)
  result = blk.call(2)
  nil
end

foo { |x| x.to_s }  # no error
bar lambda { |x| x.to_s } # error
```

egonSchiele:
I can reproduce

egonSchiele:
This is fixed in v0.15.

PR #251, opened 2017-02-20
fixes issue #242 - contract's return value is now enforced with blocks properly

esse:
This commit fixes issue with contracts for function that takes blocks. Return type is now enforced properly for blocks.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ group :test do
   gem "rspec"
   gem "aruba"
   gem "cucumber", "~> 1.3.20"
-  gem "rubocop", "~> 0.29.1", :platform => [:ruby_20, :ruby_21, :ruby_22, :ruby_23]
+  gem "rubocop", "~> 0.29.1" if RUBY_VERSION >= "2"
 end
 
 group :development do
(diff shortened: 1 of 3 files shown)


esse:
@egonSchiele do `contracts_spec_2.0.rb` spec should pass for all ruby versions? or is it specific to some type?

egonSchiele:
@esse they are specs for all rubies >= v 2.0. The logic to load that file lives [here](https://github.com/egonSchiele/contracts.ruby/blob/master/spec/spec_helper.rb#L26). So it is a legitimate test failure. Definitely one of the toughest parts of this codebase!

esse:
@egonSchiele I run `rake` (as README encouraged) instead of `rspec` (as run by travis). And `rake` shows me that all tests are passing, but `rspec` show failures.

I believe it would be nice to change it in a README - I will propose pull request with that.

esse:
@egonSchiele I commited fixes to that issue. Hoewever travis has some problems with resolving gems for Ruby 1.8 and 1.9.

egonSchiele:
Nice work, thank you!

PR #254, opened 2017-04-07
Fix StrictHash contract for extra keys

smt116:
`Contracts::StrictHash` didn't complain about extra entries in a given
hash. This was because of the missing assertion for keys. Specs hadn't
caught this case because `age` key had a wrong type anyway.

Compare keys for contract and a given hash and return false if they are
equal.

### DIFF
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -407,8 +407,8 @@ module Contracts
         return false unless arg.is_a?(Hash)
         return false unless arg.keys.sort.eql?(contract_hash.keys.sort)
 
-        contract_hash.all? do |key, _v|
-          contract_hash.key?(key) && Contract.valid?(arg[key], contract_hash[key])
+        contract_hash.all? do |key, contract|
+          contract_hash.key?(key) && Contract.valid?(arg[key], contract)
         end
       end
     end
--- a/lib/contracts/builtin_contracts.rb
+++ b/lib/contracts/builtin_contracts.rb
@@ -405,6 +405,7 @@ module Contracts
 
       def valid?(arg)
         return false unless arg.is_a?(Hash)
+        return false unless arg.keys.sort.eql?(contract_hash.keys.sort)
 
         contract_hash.all? do |key, _v|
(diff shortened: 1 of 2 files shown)


lib/contracts/builtin_contracts.rb
```diff
@@ -405,6 +405,7 @@ def initialize(contract_hash)
 
       def valid?(arg)
         return false unless arg.is_a?(Hash)
+        return false unless arg.keys.sort.eql?(contract_hash.keys.sort)
```
smt116: `sort` is missing for ruby 1.8.7, but is there any point of supporting this version? Also what do you think about comparing arrays sizes instead?

lib/contracts/builtin_contracts.rb
```diff
@@ -405,6 +405,7 @@ def initialize(contract_hash)
...
 
         contract_hash.all? do |key, _v|
           contract_hash.key?(key) && Contract.valid?(arg[key], contract_hash[key])
```
smt116: is there any point of accessing `contract_hash[key]` instead of using `_v` from the iterator? (which should be renamed to `contract` in such case)

egonSchiele:
Thanks for this PR! I will check it shortly

lib/contracts/builtin_contracts.rb
egonSchiele: nope, that sounds like a good change to me

lib/contracts/builtin_contracts.rb
egonSchiele: I think we can drop support for ruby 1.8.7, I like comparing the keys instead of array size though.

smt116:
Fixed. I didn't touched 1.8.7 support as I saw that someone else wanted to do that (https://github.com/egonSchiele/contracts.ruby/pull/255#issuecomment-294505713).

egonSchiele:
can you merge master into this branch? ci should pass, and then I'll merge

smt116:
@egonSchiele done.

egonSchiele:
Thank you!

PR #255, opened 2017-04-17
Add a `Contracts::Attrs` module containing attribute w/ contracts utilities.

frewsxcv:
This pull request adds a new module `Contracts::Attrs` containing three methods to assist in using contracts with object attributes:

- `attr_reader_with_contract <symbol>..., <contract>`
  - Wraps `attr_reader`, validates contract upon 'getting'
- `attr_writer_with_contract <symbol>..., <contract>`
  - Wraps `attr_writer`, validates contract upon 'setting'
- `attr_accessor_with_contract <symbol>..., <contract>`
  - Wraps `attr_accessor`, validates contract upon 'getting' or 'setting'

### DIFF
--- a/lib/contracts/attrs.rb
+++ b/lib/contracts/attrs.rb
@@ -1,0 +1,20 @@
+module Contracts
+  module Attrs
+    def attr_reader_with_contract(*names, contract)
+      Contract Contracts::None => contract
+      attr_reader(*names)
+    end
+
+    def attr_writer_with_contract(*names, contract)
+      Contract contract => contract
+      attr_writer(*names)
+    end
+
+    def attr_accessor_with_contract(*names, contract)
+      attr_reader_with_contract(*names, contract)
+      attr_writer_with_contract(*names, contract)
+    end
+  end
+
+  include Attrs
+end
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -606,6 +606,32 @@ Possible validator overrides:
 
 Default validators can be found here: [lib/contracts/validators.rb](https://github.com/egonSchiele/contracts.ruby/blob/master/lib/contracts/validators.rb).
 
+## Contracts with attributes
+
+You can include the `Contracts::Attrs` module in your class/module to get access to attribute utilities:
+
+- `attr_reader_with_contract <symbol>..., <contract>`
+  - Wraps `attr_reader`, validates contract upon 'getting'
+- `attr_writer_with_contract <symbol>..., <contract>`
+  - Wraps `attr_writer`, validates contract upon 'setting'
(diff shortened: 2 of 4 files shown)


frewsxcv:
On a side note: thanks for making and maintaining contracts.ruby, it's an awesome project! 👌

frewsxcv:
@egonSchiele Any thoughts about dropping support for Ruby 1.8? It's been EOL and hasn't received security updates for a few years, so to me it feels like a safe decision. If you're onboard with that, I can open a new PR for that and then rebase this on top of that.

egonSchiele:
I like this 👍 
I agree that we can drop support for Ruby 1.8.

frewsxcv:
PR opened: https://github.com/egonSchiele/contracts.ruby/pull/256

egonSchiele:
can you merge master into this branch? ci should pass, and then I'll merge

frewsxcv:
Rebased.

frewsxcv:
Is that error related to my change here? https://github.com/egonSchiele/contracts.ruby/commit/8e6d4eec01bc64bff50192a63ac236ee34449401#diff-eacde0c7f8149b55a880ae5fdcd84b3aL7

egonSchiele:
Looks like the issue is, methods created using `attr_*` don't have a `source_location`, which makes sense since they are generated dynamically. This diff fixes ci:

```
diff --git a/lib/contracts/support.rb b/lib/contracts/support.rb
index 780455c..f3efa20 100644
--- a/lib/contracts/support.rb
+++ b/lib/contracts/support.rb
@@ -5,7 +5,11 @@ module Contracts
         return method.method_position if method.is_a?(MethodReference)

         file, line = method.source_location
-        file + ":" + line.to_s
+        if file.nil? || line.nil?
+          ""
+        else
+          file + ":" + line.to_s
+        end
       end

       def method_name(method)
```

I can merge this branch and fix.

egonSchiele:
thank you!

frewsxcv:
More inportantly, thanks for maintaining this great library :)

lib/contracts/attrs.rb
```diff
@@ -0,0 +1,20 @@
...
+      attr_writer(*names)
+    end
+
+    def attr_accessor_with_contract(*names, contract)
+      attr_reader_with_contract(*names, contract)
+      attr_writer_with_contract(*names, contract)
+    end
+  end
+
+  include Attrs
```
PikachuEXE: Why this is included in place?
Allow people to just include ::Contract & use all methods?

egonSchiele: good point

frewsxcv: i don't remember why i added that line; feel free to remove it

PR #259, opened 2017-05-22
Apply contracts to each name for multi-attribute calls

kyeah:
:wave:

When passing in multiple fields to a single attr_x_with_contract call, the current implementation only applies the contract for the first field.

```ruby
class Thinger
  include Contracts
  include Contracts::Attrs

  attr_accessor_with_contract :a, :b, :c, String
end
```
```
[development] » t = Thinger.new
» #<Test:0x007f934f552dc0>
[development] » t.c
» nil
[development] » t.b
» nil
[development] » t.a
ReturnContractError: Contract violation for return value:
Expected: String,
        Actual: nil
        Value guarded in: Test::a
```

This applies the contract and attr_reader individually to each field.

### DIFF
--- a/spec/attrs_spec.rb
+++ b/spec/attrs_spec.rb
@@ -9,11 +9,15 @@ RSpec.describe "Contracts:" do
         @name_r = name
         @name_w = name
         @name_rw = name
+
+        @name_r_2 = name
+        @name_w_2 = name
+        @name_rw_2 = name
       end
 
-      attr_reader_with_contract :name_r, String
-      attr_writer_with_contract :name_w, String
-      attr_accessor_with_contract :name_rw, String
+      attr_reader_with_contract :name_r, :name_r_2, String
+      attr_writer_with_contract :name_w, :name_w_2, String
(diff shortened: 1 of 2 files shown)


spec/attrs_spec.rb
```diff
@@ -27,6 +31,16 @@ def initialize(name)
...
       end
 
+      it "getting valid type for second val" do
+        expect(Person.new("bob").name_r_2)
+          .to(eq("bob"))
+      end
+
+      it "getting invalid type for second val" do
+        expect { Person.new(1.3).name_r_2 }
+        .to(raise_error(ReturnContractError))
```
egonSchiele: Rubocop expects this to be indented two spaces: https://travis-ci.org/egonSchiele/contracts.ruby/jobs/234892156

kyeah: fixed in 7721271 👍

kyeah:
Opened #260 due to the failing 1.9.2 ci check.

egonSchiele:
Thank you!

PR #270, opened 2017-09-24
RuboCop: Upgrade to 0.41.2, last to support Ruby1.9

olleolleolle:
This PR:

- updates RuboCop to 0.41.2, last to support Ruby1.9
- sets TargetRubyVersion to `1.9`
- sets a RuboCop exclusion on all the spec files which were Ruby-version-specific
- generates and includes a RuboCop TODO file with all the current lint failures

This sets us off on the journey of upgrading RuboCop and keeping up-to-date with style.


olleolleolle:
One way of making this less irritating for the CI matrix would be to make the RuboCop check a "Build Stage" (See https://docs.travis-ci.com/user/build-stages/ for more), with a single Ruby version in it that stage.

PR #272, opened 2017-10-11
Fix typo in Contracts::MethodHandler#validate_decorators!

indigoviolet:
AFAICT, `name` is not defined at that point


egonSchiele:
Thanks!

PR #273, opened 2017-10-12
RFC: Flags to turn off pattern matching and evaluation

indigoviolet:
We'd like to be able to turn off the pattern matching/method overloading behavior altogether -- it can lead to unpleasant surprises, and we don't want someone inadvertently writing code that relies on it in our environment.

Similarly, I think it'd be valuable to have a feature that allows selective disabling of *validation* of contracts after having loaded them. (The idea being, it might permit running contracts in non-perf-sensitive codepaths in production or a fraction of prod requests)

I've sketched a version of how I thought this might work to get some feedback from you about

a) is this likely to get merged if completed?
b) is this a reasonable approach?
c) any big pitfalls that I'm not thinking of

I haven't tested this very much at all, will proceed to do so if you think this is a good direction to pursue.

### DIFF
--- a/lib/contracts/support.rb
+++ b/lib/contracts/support.rb
@@ -12,6 +12,22 @@ module Contracts
         end
       end
 
+      def decorators_enabled?
+        !ENV["NO_CONTRACTS"]
+      end
+
+      def invariants_enabled?
+        !ENV["NO_CONTRACTS"]
+      end
+
+      def pattern_matching_enabled?
+        !ENV["NO_CONTRACTS_PATTERN_MATCHING"]
+      end
+
+      def validation_enabled?
+        !ENV["NO_CONTRACTS_VALIDATION"]
+      end
+
       def method_name(method)
         method.is_a?(Proc) ? "Proc" : method.name
       end
--- a/lib/contracts/call_with.rb
+++ b/lib/contracts/call_with.rb
@@ -9,46 +9,12 @@ module Contracts
       # Explicitly append options={} if Hash contract is present
       maybe_append_options!(args, blk)
 
-      # Loop forward validating the arguments up to the splat (if there is one)
-      (@args_contract_index || args.size).times do |i|
-        contract = args_contracts[i]
-        arg = args[i]
-        validator = @args_validators[i]
-
-        unless validator && validator[arg]
-          return unless Contract.failure_callback(:arg => arg,
-                                                  :contract => contract,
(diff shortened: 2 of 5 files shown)


lib/contracts/support.rb
```diff
@@ -12,6 +12,22 @@ def method_position(method)
         end
       end
 
+      def decorators_enabled?
+        !ENV["NO_CONTRACTS"]
+      end
+
+      def invariants_enabled?
+        !ENV["NO_CONTRACTS"]
```
indigoviolet: this should probably be controlled along through `NO_CONTRACTS_VALIDATION`?

egonSchiele:
I like this idea! Answers:

a) is this likely to get merged if completed? *Yes.*
b) is this a reasonable approach? *Yes.*
c) any big pitfalls that I'm not thinking of *Here are some things that have caused issues in the past, I would construct a test case with a combination of these: pattern matching (especially with  calls that cause contract failures) + contracts in modules instead of classes, with those modules being then included in a class + contracts inside a `class << self` block.

I like the support class, and I like the option to turn off decorators, invariants, and pattern matching selectively. I don't understand:

> Similarly, I think it'd be valuable to have a feature that allows selective disabling of validation of contracts after having loaded them. (The idea being, it might permit running contracts in non-perf-sensitive codepaths in production or a fraction of prod requests)

Isn't this essentially the existing `NO_CONTRACTS` functionality?

indigoviolet:
> Isn't this essentially the existing `NO_CONTRACTS` functionality?

Maybe I'm misunderstanding, but my reading of the code and experimentation suggests not-

https://github.com/egonSchiele/contracts.ruby/blob/78be9e20218d60f4619d5ea1a16fc7892064f500/lib/contracts/method_handler.rb#L102

We will completely ignore decorators at startup if `NO_CONTRACTS` is set; I'm interested in something more dynamic, where I can write a block similar to

https://github.com/egonSchiele/contracts.ruby/blob/78205dee5df85bb1f75a448552be583cd8373c90/spec/support.rb#L2

but for use in non-test code- ie. decorators are defined but not evaluated.

This probably shouldn't be done via environment variable, and instead some global variable -- I'll explore that.

egonSchiele:
Oh that's a neat idea. Agreed, NO_CONTRACTS doesn't let you suspend contracts for a short time

nathan-appere:
@indigoviolet did you end up using this in prod?

Issue #278, opened 2017-12-12
keep properties of given blocks


md-work:
Since #242 (Func contract's return value isn't enforced with blocks) was fixed in contracts-0.15, given blocks are being put into a lambda. So the forwarded block isn't 100% identical with the original one. Especially it's `arity` and it's `source_location` changed.

Please keep as many properties as possible of the original block, especially the arity.

Or find a way to forward the original block without breaking #242 again. So the original block could be kept. I'd really appreciate any ideas for this, because it might be the cleanest solution!

&nbsp;

---

&nbsp;

Commit for #242, introducing this bug: https://github.com/egonSchiele/contracts.ruby/commit/13e56bdf859a54c569d1a7fe4680251880c6d089#diff-a45dd775c2769b9ab16a23f9d4560adc
`lib/contracts/call_with.rb`
```
@@ -73,7 +75,8 @@ module Contracts
                  method.call(*args, &blk)
                else
                  # original method name referrence
-                 method.send_to(this, *args, &blk)
+                 added_block = blk ? lambda { |*params| blk.call(*params) } : nil
+                 method.send_to(this, *args, &added_block)
                end
 
       unless @ret_validator[result]
```

&nbsp;

---

&nbsp;

I especially need the arity to stay unchanged. Some examples about arity:
```
def a(&b)
  puts b.arity
end

a {|x| puts x}
=> 1

a {|x,y| puts x+y}
=> 2

a {puts ''}
=> 0

a {|*x| puts x}                                                                                                   
=> -1                                                                                                                                   
```

&nbsp;

A very "simple" way how to keep the arity is using eval (see below). But I hope we find a better way.
```
--- a/lib/contracts/call_with.rb
+++ b/lib/contracts/call_with.rb
@@ -81,7 +81,15 @@ module Contracts
                  method.call(*args, &blk)
                else
                  # original method name reference
-                 added_block = blk ? lambda { |*params| blk.call(*params) } : nil
+                 added_block = nil
+                 if blk
+                    params_ary = []
+                    for i in 1..blk.arity
+                      params_ary << 'x_'+i.to_s
+                    end
+                    params_str = params_ary.join(',')
+                    added_block = eval('lambda { |'+params_str+'| blk.call('+params_str+') }')
+                 end
                  method.send_to(this, *args, &added_block)
                end
```

md-work:
Looks like every Block being passed is wrapped into a lambda. Even if there's no "subcontract" checking the blocks parameters.

[...]

PR #281, opened 2017-12-14
Wrapping &blocks only when there is a Func check. (bug 278)

md-work:
See issue 278: https://github.com/egonSchiele/contracts.ruby/issues/278#issuecomment-351428779

### DIFF
--- a/lib/contracts/call_with.rb
+++ b/lib/contracts/call_with.rb
@@ -81,8 +81,8 @@ module Contracts
                  method.call(*args, &blk)
                else
                  # original method name reference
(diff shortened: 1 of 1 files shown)


egonSchiele:
[This is causing infinite recursion](https://travis-ci.org/egonSchiele/contracts.ruby/jobs/316352384#L603), maybe because of the `is_a?` call?

md-work:
Fixed the problem in a new request (I couldn't update this pull request):
https://github.com/egonSchiele/contracts.ruby/pull/283

Somehow Travis CI didn't like that I reassigned the blk variable.

PR #283, opened 2018-02-05
Wrapping &blocks only when there is a Func check. (second version) (bug 278)

md-work:
See https://github.com/egonSchiele/contracts.ruby/issues/278#issuecomment-351428779 and https://github.com/egonSchiele/contracts.ruby/pull/281#issuecomment-362851436

Somehow Travis CI didn't like that I reasigned the blk variable. So I introduced a new target_blk variable.


egonSchiele:
Thanks!

PR #285, opened 2018-03-14
Update README.md to reflect new ruby versions.

victormartins:
Changing the documentation since we now test for Ruby versions 2.3, 2.4 and 2.5.


egonSchiele:
Thanks!

PR #287, opened 2018-09-05
Fix potentially incorrect example in feature file.

frewsxcv:
`1 - 1` is not a positive natural number


PikachuEXE:
Oops no one check this for 6 years (I am a new helper here
The point of returning `0` on inputing `1` is exactly the point to test the return contract error is raised
See `Scenario: Rejects negative integers as a return value`

If you meant to do something else please reply

PR #288, opened 2018-10-04
Update gemspec to only allow lowest tested version

### LINKED ISSUE #275: Specify minimum Ruby version in the gemspec
PR #256 dropped support for Ruby 1.8. However, this is not enforced in the gemspec. This means contracts 0.16.0 will install fine on 1.8, only to fail at runtime. Please consider setting `required_ruby_version` there.

WA9ACE:
Fixes #275


WA9ACE:
Fixes #275

PR #298, opened 2022-01-24
Add Ruby 3.1 to the CI matrix

petergoldstein:
This PR just adds Ruby 3.1 to the CI matrix.

As a side note, there's an `include` in the config that I don't think is necessary.  The specs run twice for Ruby 3.0 - once with both rspec and cucumber, and then one with rspec only.  The latter is the result of the include, which I believe can be removed.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -22,6 +22,7 @@ jobs:
         os:
           - ubuntu
         ruby:
+          - "3.1"
           - "3.0"
         test_command:
           - "bundle exec rspec && bundle exec cucumber"


PikachuEXE:
You are right
Before it's like [this](https://github.com/egonSchiele/contracts.ruby/commit/e7a9ef4b2ac9ec26bd55c54c113474d60235bdc6#diff-95557d53bf91069e59d82b9d8fcfaf52ec84a762858983edd5ef3c9b3e4c8191)
But I still have no idea why that `include` is needed even before the change

PikachuEXE:
Oops since I updated the config now we have code conflict 😅

PikachuEXE:
Just force push without closing is fine

petergoldstein:
Ok, that should work.  Had a merge issue.

PikachuEXE:
Thanks :)

PR #306, opened 2024-09-22
Always load version.rb, suppress legacy deprecation warning

vlad-pisanov:
Two minor improvements:

1. Automatically load `version.rb`, so that `Contracts::VERSION` is always accessible to user
2. Hide the gem deprecation warning if the user is already on the latest version

### DIFF
--- a/contracts.gemspec
+++ b/contracts.gemspec
@@ -13,8 +13,4 @@ Gem::Specification.new do |s|
   s.homepage    = "https://github.com/egonSchiele/contracts.ruby"
   s.license     = "BSD-2-Clause"
   s.required_ruby_version = [">= 3.0", "< 4"]
-  s.post_install_message = "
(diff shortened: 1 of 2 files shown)


contracts.gemspec
```diff
@@ -13,8 +13,11 @@ Gem::Specification.new do |s|
...
-    0.16.x will be the supporting Ruby 2.x and be feature frozen (only fixes will be released)
-    For Ruby 3.x use 0.17.x or later
-  "
+
+  if Contracts::VERSION < "0.17"
+    s.post_install_message = "
+      0.16.x will be the supporting Ruby 2.x and be feature frozen (only fixes will be released)
+      For Ruby 3.x use 0.17.x or later
+    "
+  end
```
PikachuEXE, line 22: Might as well just remove the whole message for 0.17+ = `master`

PikachuEXE (approved):

